PR #2, opened 2018-05-03
WIP: Add formatting support.

### LINKED ISSUE #1: Implement formatting support
I started work on this. I thought it would be rather straight forward using the pretty printer. But that doesn't reindent at all.

dvc94ch:
Seems to work, may need some more testing.

Closes #1

### DIFF
--- a/format.rkt
+++ b/format.rkt
@@ -1,0 +1,66 @@
+#lang br
+(require br/indent
+         framework
+         racket/contract/base
+         rackunit)
+
+;; Procedure to be used by lang-server
+(define (format text
+                #:start [start 0]
+                #:end [end (string-length text)]
+                #:tab-size [tab-size 2])
+  (define (str->text str)
+    (define t (new racket:text%))
+    (send t insert-port (open-input-string str))
+    t)
+  (define t (str->text text))
+  (define start-line (line t start))
+  (define end-line (line t end))
+  (define start-pos (line-start t start-line))
+  (define end-pos (add1 (line-end t end-line)))
+  (define delta 1)
+  (define indented-t
+    (for/fold ([t-acc t])
+        ([line-idx (in-range start-line (add1 end-line))])
+      ;; simulate DrR indentation
+      ;; by dropping leading spaces and applying new indent.
+      (define line-start-pos (line-start t-acc line-idx))
+      ;; add1 to grab ending newline too
+      (define line-end-pos (add1 (line-end t-acc line-idx)))
+      (define new-indent (or (send t-acc compute-amount-to-indent line-start-pos) 0))
+      (define new-line-str
+        (list->string (append (make-list new-indent #\space)
+                              (dropf (line-chars t-acc line-idx) space-char?))))
+      (send t-acc delete line-start-pos line-end-pos)
+      (send t-acc insert new-line-str line-start-pos)
+      ;; compute delta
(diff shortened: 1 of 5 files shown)


cfinegan:
Hi David,

Thank you for your contribution! I'm glad you're interested in working on racket-langserver. 

I'm not sure that re-implementing code formatting in a way that mimics DrRacket's code formatting is the best way to go about this. Ideally we'd like the language server invoke functionality that already exist in the drracket package (you can think of the language server as a "headless mode" for DrRacket). 

Your implementation seems to have some inconsistencies when compared to DrRacket's code formatting. For example, using `append-message.rkt` from this very project as a test, your formatter generates result different from the DrRacket formatter on lines 3, 4, 16, 23, and 24 (the indentation of those lines seem to be off by one compared to DrRacket). This may be an easy bug to individually fix, but more generally I would hope that we wouldn't have to re-implement a code formatter that mirrors the DrRacket formatter. This is extra work and an additional maintenance burden going forward.

That being said, it might still make sense to pull this in as  WIP, since it's much better than the nothing that we currently have. Thoughts, @jeapostrophe ?

Either way, it's hard to test this functionality in Atom unless `atom-ide-racket` is set up to forward code format requests to the langserver (which it currently does not do). I'm going to push an update to the `atom-ide-racket` to rectify this, so it should be easier to test this functionality going forward. 

One other minor nitpick: If you're going to use `#lang br`, then you should add `beautiful-racket` to the list of dependencies in `info.rkt`. I got a runtime error when I first ran your branch and had to do some googling to figure out what dependency was missing.

Thanks again,
Conor Finegan

cfinegan:
Specifically, I think we should use [drracket:indentation](https://docs.racket-lang.org/tools/lang-languages-customization.html?q=indentation#(idx._(gentag._9._(lib._scribblings%2Ftools%2Ftools..scrbl)))). This should give us the same results as DrRacket and we already use a text% object so it seems pretty straightforward.

dvc94ch:
I think that [`compute-amount-to-indent`](https://github.com/racket/drracket/blob/bdd57e178d0ae11f37ca65499212759d3da6659e/drracket/drracket/private/module-language-tools.rkt#L490-L496) already calls out to `drracket:indentation`. Searching the code base for `compute-amount-to-indent` doesn't yield any results, so I'm stomped as to where this procedure is used in dr. racket.

cfinegan:
I'm not sure that it is used by DrRacket. Referencing [this article on how DrRacket pares #lang-based languages](https://docs.racket-lang.org/tools/lang-languages-customization.html), it seems that DrRacket uses `read-language` to get the current language's *get-info* procedure, then queries that procedure with the value `'drracket:indentation` to get the procedure used to indent lines. 

The fact that using `compute-amount-to-indent` gives different indentation results than DrRacket suggests that DrRacket isn't using `compute-amount-to-indent` internally, or at least it seems that way to me. I am also unable to find any examples of `compute-amount-to-indent` (or `compute-racket-amount-to-indent`) being invoked from within the DrRacket code base, so it seems to me like this may be dead code? It's hard to for me to say, as the object-oriented Racket code is very spread out and I find it difficult to parse.

dvc94ch:
I'll be cleaning my PR up and resubmitting soon:
Server side coloring and indentation:
screenshot from 2018-05-13 10-51-38
It's kind of unbearably slow at the moment...

PR #7, opened 2018-05-22
fix macro for new syntax form

rmculpepper:
Racket's syntax form now interprets ??, so that identifier
should be quoted (or escaped) so that it gets interpreted as
part of the inner template.


jeapostrophe:
Thank you!

PR #10, opened 2019-08-07
Add specialized logic for Windows resolution of file:// URIs

dergemkr:
Note: I don't have access to a non-Windows environment for testing that this maintains current functionality in other situations. Please double-check.


jeapostrophe:
Thanks!

PR #11, opened 2020-04-10
fix textDocument/documentSymbol

JJPro:
I made **symbol lookup**/**document outline** working. 
Previously it would list all words in the document and the server would crash for big documents, 
Now it only looks for function definitions (`define`) and type definitions (`define-type`)


jeapostrophe:
I don't think this is a good change, because it is brittle in the face of macros that perform binding. The existing approach is to give all symbols and allow the lsp to basically be like emacs' `M-/` where it just helps you complete symbols that already there. A more robust change would be to do what DrRacket does and look at the expansion for binding positions.

Eugleo:
Do we have any way to find out which expressions are binding, without resorting to some sort of string pattern match?

jeapostrophe:
DrRacket has a library implement for this, as part of Check Syntax, that fully-expands the program and then inspects the source code for binding information in the syntax objects.

text-document.rkt
```diff
@@ -277,9 +280,15 @@
...
-  (if (procedure? lexer) ;; TODO: Is this an issue with module-lexer docs?
-      lexer
-      (error 'get-lexer "~v" lexer)))
+  #|
+    lexer is one of:
+      - 'before-lang-line
+      - lexer, the procedure
+    loop until it is a procedure
+  |#
+  (if (procedure? lexer)
```
Eugleo, line 289: This I think is problematic, as for empty files it seems to cycle indefinitely. It's getting too late here, but I'll push a fix tomorrow. We should also properly detect the defining, as @jeapostrophe suggested.

Eugleo: As per `module-lexer` [docs](https://docs.racket-lang.org/syntax-color/index.html?q=module-lexer#%28def._%28%28lib._syntax-color%2Fmodule-lexer..rkt%29._module-lexer%29%29):

> When mode is #f (indicating the start of the stream), the lexer checks in for a #lang specification.
> 
> If a #lang line is present but the specified language does not exist, the entire in input is consumed and colored as 'error.
> 
> If the language exists and the language provides a get-info function, then it is called with 'color-lexer. If the result is not #f, then it should be a lexer function for use with color:text<%>. The result mode is the lexer—paired with #f if the lexer is a procedure arity 3—so that future calls will dispatch to the language-supplied lexer.
> 
> If the language is specified but it provides no get-info or 'color-lexer result, then racket-lexer is returned as the mode.

When the file is empty, I'd say `module-lexer` will plaily return `#f`, among other things — and calling it again and again won't change a thing.

Eugleo: Question for @jeapostrophe: the default behaviour seems to be to error out if a lexer procedure isn't returned. I'm not familiar with your codebase nor with LSP too much, but it doesn't seem right to give errors even in non-erroneous state (= plain empty file).

jeapostrophe: An empty file is definitely an error in Racket, because a `#lang` or `(module ...)` is required.

Eugleo: That makes sense, thanks.

PR #15, opened 2020-05-02
Fix infinite loop caused by document-symbol

Eugleo:
Your version of `get-lexer` caused infinite loop in empty files. This fixes it, please merge ASAP.


jeapostrophe:
We definitely don't want to infinite loop... but it should be some error.

PR #20, opened 2020-09-22
Produce a best-effort diagnostic for missing module with no src loc

dalev:
The language server reports an uncaught exception if you point it at a module that contains a bogus require, like:
```
#lang racket
(require bogus/module/path)
```
This patch catches the exception and turns it into a diagnostic message.  

Fwiw, DrRacket fares no better: it also lacks the source location information.  It does not highlight the offending require, and it just displays the exn message at the bottom.  

Unfortunately, the LSP appears to require a source location, so I punted and just blamed the first character of the module.  Otherwise, this lang server behavior is now consistent with DrR.

### DIFF
--- a/check-syntax.rkt
+++ b/check-syntax.rkt
@@ -80,15 +80,30 @@
 
 (define ((error-diagnostics src) exn)
   (define msg (exn-message exn))
-  (define get-srclocs (exn:srclocs-accessor exn))
-  (define srclocs (get-srclocs exn))
-  (for/list ([sl (in-list srclocs)])
-    (match-define (srcloc src line col pos span) sl)
-    (Diagnostic #:range (Range #:start (Pos #:line (sub1 line) #:char col)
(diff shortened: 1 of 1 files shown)


jeapostrophe:
Thanks @dalev 

This is interesting, because the documentation says that "The default module name resolver raises this exception only when it is not given a syntax object as its second argument, and the default load handler raises this exception only when the value of current-module-path-for-load is not a syntax object." --- https://docs.racket-lang.org/reference/exns.html#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._exn~3afail~3afilesystem~3amissing-module%29%29

It is surprising that it's getting to there with no syntax object. Do you have any idea @rfindler?

rfindler:
I'm not sure. I do believe it would be great if we could somehow communicate the source location of a `require` that triggered the filesystem error that eventually happens when an `open-file` fails.

PR #22, opened 2020-10-19
Use racket-lexer if no lang line presented.

shih-liang:
If no #lang presented, use racket lexer(syntax-color/racket-lexer).


jeapostrophe:
Good idea!

PR #24, opened 2020-12-21
Add formatting support

### LINKED ISSUE #1: Implement formatting support

MLavrentyev:
This adds formatting support. Gets the indenter using `read-language` and getting the `drracket:indentation` setting (with a default fallback to use the racket indenter).

Making the draft pr now so it's visible. Will be testing in the near future:
 - #lang with different indenter

Closes #1 .

### DIFF
--- a/check-syntax.rkt
+++ b/check-syntax.rkt
@@ -80,7 +80,7 @@
 
 (define ((error-diagnostics src) exn)
   (define msg (exn-message exn))
-  (cond 
+  (cond
     [(exn:srclocs? exn)
      (define srclocs ((exn:srclocs-accessor exn) exn))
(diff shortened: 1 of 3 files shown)


jeapostrophe:
This looks like the right thing! I'm glad it was relatively easy to do. Let me know when you've done some more testing and think it is ready to go! Thank you!

MLavrentyev:
Ok seems to be good to go - it's succesfully getting the right #lang indenter for the given #lang and indenting correctly for the few custom indenters I've tried.

jeapostrophe:
Awesome, thank you!

PR #25, opened 2020-12-24
Fix indenter non-idempotence bug

MLavrentyev:
As I was starting to use the indenter, I ran into a bug where the versions of the document in the client (i.e. in the buffer) vs. the server were getting to be mismatched, resulting in some weird back-and-forth indenting (whereas formatting should be idempotent).

Turns out, language servers (vscode for sure) send an change notification after the indenting is done in the buffer, giving us double edits in the server version of the doc. 

This pr fixes it by indenting over a dummy doc-text called `mut-doc-text` (this is necessary since the indenter relies on previous lines).

Note: only the last two commits are new. For some reason, the fork is comparing all of them, though.

### DIFF
--- a/check-syntax.rkt
+++ b/check-syntax.rkt
@@ -80,7 +80,7 @@
 
 (define ((error-diagnostics src) exn)
   (define msg (exn-message exn))
-  (cond 
+  (cond
     [(exn:srclocs? exn)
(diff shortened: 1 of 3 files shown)


jeapostrophe:
I think the same reason that Github is showing multiple commits is preventing me from merge. Can you rebase it on your end?

Issue #21, opened 2020-10-04
Integrate with vim(vim-lsp)


gn0uchat:
Want to integrated with vim/vim-lsp. 
I have no idea how to use this package.
May I ask where to start?

jeapostrophe:
You install the package with `raco pkg`, then set `racket -l racket-langserver` as the LSP command.

Lazerbeak12345:
This command is also useful for `nvim-lspconfig` perhaps this command could be documented in the readme?

Lazerbeak12345:
Interesting note: when I run that command in my terminal, here's what I get:

```racket -l racket-langserver
/usr/racket/share/pkgs/gui-lib/framework/private/preferences.rkt:581:19: string-
constant: autoload-automatically-reload is not a known string constant
  in: (string-constant autoload-automatically-reload)
  location...:
   /usr/racket/share/pkgs/gui-lib/framework/private/preferences.rkt:581:19
  context...:
   do-raise-syntax-error
   /usr/racket/share/pkgs/string-constants-lib/string-constants/string-constant.
rkt:228:0: string-constant
   apply-transformer-in-context
   apply-transformer52
   dispatch-transformer41
   for-loop
   [repeats 2 more times]
   loop
   [repeats 3 more times]
   finish-expanding-body33
   lambda-clause-expander
   for-loop
   [repeats 1 more time]
   finish-bodys
   lambda-clause-expander
   loop
   ...
```
I believe this is preventing it from working at all. Should I make a seperate issue for this? (I'd be happy to) Or perhaps did I miss a configuration step?

jeapostrophe:
That problem is a mismatch between the `string-constants` package and the `gui-lib` package. My guess is that on your Racket 7.5, just one of them got updated, but in your Racket 7.8, they were aligned.

weskerfoot:
> You install the package with `raco pkg`, then set `racket -l racket-langserver` as the LSP command.

How?

PR #27, opened 2020-12-29
Add details on how to use this LSP for non-Atom editors

jo-sm:
This PR clarifies how to setup `racket-langserver` for IDEs that don't necessarily have a package for it.

I recently setup `racket-langserver` in Sublime after looking up how to add a custom client and following the simple instructions in https://github.com/jeapostrophe/racket-langserver/issues/21#issuecomment-703287071, so I figure it makes sense to put them in the README for others to more easily see.

I also moved the code formatting section into *Currently Supported* since as far as I can tell support has been added now (https://github.com/jeapostrophe/racket-langserver/pull/24). 

Do y'all have any thoughts on a troubleshooting or additional info to add here, and if so, what would go in there? Is there a minimum version you should have for the package? My Racket is at 7.9 but I saw in the comment linked above someone may have issues if they have an older version, so it might be good to add to the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,32 @@
 # racket-langserver
 
-[Language Server Protocol](http://langserver.org/) implementation for Racket. This project seeks to use DrRacket's public APIs to provide functionality that mimics DrRacket's code tools as closely as possible.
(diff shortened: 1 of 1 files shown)


jeapostrophe:
This is great. Thank you. I don't believe that there are any big version pre-reqs, because I think we're relying on fairly "old" Racket APIs. If people find examples of combinations that are broken, then the best thing to do is add a requirement in `info.rkt` at the Racket level and then `raco` will complain. Better still we could add version exceptions to the package server so that `raco` will find the correct version.

PR #33, opened 2021-03-28
Display a link to docs in textDocument/hover

Keterr:
If docs are found, provide a link after the hover text. A sufficient way of bringing DrRacket's "View documentation for..." functionality to other editors.

image

The link needs to be converted from filepath to equivalent URL or else editors may open the HTML document in-editor instead of opening it in a new browser - Visual Studio Code exhibits this behavior. So it will fail on library documentation not hosted on docs.racket-lang.org.

### DIFF
--- a/check-syntax.rkt
+++ b/check-syntax.rkt
@@ -6,6 +6,7 @@
          racket/gui/base
          racket/match
          racket/set
+         net/url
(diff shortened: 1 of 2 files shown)


jeapostrophe:
Cool! Thanks

PR #34, opened 2021-03-28
replace CRLF with LF at read-time

### LINKED ISSUE #32: CRLF line endings handled incorrectly
This bug causes CRLF line endings to be double-counted, causing line positions to be reported incorrectly. The result is that generated symbols are entirely useless and the langserver is virtually unusable for editors configured to use CRLF.

It is caused by a bug in framework:
image 
https://github.com/racket/gui/issues/221

As I don't anticipate it to be fixed there any time soon a temporary fix could be considered here, such as manually replacing CRLF with LF before sending off to the `racket:text` instance.

Keterr:
Fixes #32 

Not an ideal solution but it seems to work just fine.

### DIFF
--- a/msg-io.rkt
+++ b/msg-io.rkt
@@ -2,6 +2,7 @@
 (require json
          racket/contract/base
          racket/match
+         racket/string
          (only-in racket/port open-output-nowhere))
 
 (define verbose-io? (make-parameter #f))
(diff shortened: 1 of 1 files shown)


jeapostrophe:
Thanks for the fix... gross that we need to do this, but I agree it's a nice short-term fix.

PR #35, opened 2021-03-28
Add textDocument/signatureHelp, cache traces

Runi-c:
Uses scribble/blueboxes to provide bluebox docs:

https://user-images.githubusercontent.com/5150427/112769677-88c4a600-8fdf-11eb-9888-a8bccda735d9.mp4

In order to provide this while the user is still writing, and also for a better overall experience, this PR also caches traces so that if subsequent edits cause errors the rest of the useful trace info (requires, hovers, definitions, ...) are not lost. Trace interval maps are expanded or contracted on edits so symbols are persistently tracked correctly.

In order to provide docs on symbols that have not yet been resolved (due to incomplete expressions, etc), I also moved the symbol enumeration from textDocument/symbols to `(check-syntax)` so that they can be used by other handlers.

### DIFF
--- a/check-syntax.rkt
+++ b/check-syntax.rkt
@@ -67,10 +67,10 @@
                                              (Decl (Decl-require? decl-set) (+ (car d-range) amt) (+ (cdr d-range) amt))
                                              #f)]
                                         [else 
-                                         (set-map decl-set (lambda (d-range) 
+                                         (list->set (set-map decl-set (lambda (d-range) 
                                                              (if (> (car d-range) after)
                                                                  (cons (+ (car d-range) amt) (+ (cdr d-range) amt))
-                                                                 d-range)))]))
+                                                                 d-range))))]))
                        (when result
                          (interval-map-set! int-map (car range) (cdr range) result)))))
     ;; Getters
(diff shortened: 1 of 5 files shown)


jeapostrophe:
This is cool. I see you're still making a bunch of changes... what do you think still has to be done?

Keterr:
> This is cool. I see you're still making a bunch of changes... what do you think still has to be done?

I didn't anticipate working on this PR more but since the cache change touches the behavior of every single method there was some extra bugs and unexpected behavior to squish out. I think it's in a perfectly usable state now.

Keterr:
Case in point: Didn't realize there were additional intervals being tracked as the values of some of the interval maps. Those need to be moved too in order for textDocument/definition and textDocument/references to work properly.

I *think* this PR should be bug-free now but you know how bugs are so I'd understand waiting a day or two in case I find anything else.

jeapostrophe:
I'm going to wait a little bit :) Will you ping me next week if there are no changes?

Runi-c:
@jeapostrophe I've been testing the PR for the past few days to write other features for the langserver, only noticed one elusive error which I finally sat down and fixed. I'm pretty confident this one is good to go now, don't mean to rush things but I've got a couple other PRs waiting on this one to be merged and I'd love to keep moving on this project while I still have the time and motivation lol.

jeapostrophe:
Sounds good! Thanks

PR #37, opened 2021-04-02
Provide textDocument/onTypeFormatting

Runi-c:
Formats entire s-expressions when closing blocks and moves indentation to the correct level when the user starts a new line, when "Format On Typing" is available and enabled in the user's editor.

https://user-images.githubusercontent.com/5150427/113434327-f84ee280-939d-11eb-872f-dee9383a8170.mp4

This and auto-completion are currently a bit sluggish in practice because the language server must synchronously re-check syntax right when the user types, which blocks these providers. I have an experimental multi-threading branch I'm working on now that should fix this.

### DIFF
--- a/methods.rkt
+++ b/methods.rkt
@@ -75,6 +75,8 @@
        (text-document/formatting! id params)]
       ["textDocument/rangeFormatting"
        (text-document/range-formatting! id params)]
+      ["textDocument/onTypeFormatting"
+       (text-document/on-type-formatting! id params)]
       [_
        (eprintf "invalid request for method ~v\n" method)
        (define err (format "The method ~v was not found" method))
@@ -116,7 +118,9 @@
                'documentHighlightProvider #t
                'documentSymbolProvider #t
                'documentFormattingProvider #t
-               'documentRangeFormattingProvider #t))
+               'documentRangeFormattingProvider #t
+               'documentOnTypeFormattingProvider (hasheq 'firstTriggerCharacter ")" 'moreTriggerCharacter (list "\n"))))
+
      (define resp (success-response id (hasheq 'capabilities server-capabilities)))
      (set! already-initialized? #t)
      resp]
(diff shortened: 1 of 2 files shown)


jeapostrophe:
Re: multi-threading --- I think the correct approach is to make something like a `kill-safe` server that queues the requests and if it is handling request N, but there have come in N+1, N+2, and N+3, then it ignores N+1 and N+2 and just goes to N+3

jeapostrophe:
And this is cool!

Runi-c:
> Re: multi-threading --- I think the correct approach is to make something like a `kill-safe` server that queues the requests and if it is handling request N, but there have come in N+1, N+2, and N+3, then it ignores N+1 and N+2 and just goes to N+3
Do you think I could get some pointers to relevant docs for implementing something like this? Still a relative novice to Racket so I don't know the best ways to go about running a parallel server or making it `kill-safe`. The implementation I was tinkering with so far is to keep a map of URIs to current check-syntax threads, and when a new edit comes in, simply `kill-thread`ing the old one and starting a new check syntax.

jeapostrophe:
I'm referring to this paper --- https://www.cs.utah.edu/plt/publications/pldi04-ff.pdf --- with slides --- http://www.cs.utah.edu/~mflatt/tmp/pldi-slides.pdf

By server, I just mean a `thread` that has some way of the rest of the application communicating with it via channels or other such things.

I think your idea could work well, although I think you don't want to just kill the old one always, because if someone is constantly typing you may never finish. I think instead you'd want to let the current one run and then just set a signal that it should re-run when it is done. Maybe you could do something like:

For every URI, start a thread that can see the code and a semaphore. The thread waits for the semaphore to be true and spawns a Check Syntax when it is true. When an edit comes in, it sets the semaphore, because there's a mapping from URIs to the semaphores.

This way if there are 10 edits, the first one will cause a Check Syntax to start and then the other 9 will be idempotent and just cause a single Check Syntax at the end, so there are 2 total, rather than 10, with most of them being out-of-date.

PR #39, opened 2021-04-03
Direct VSCode users to Magic Racket

jackfirth:
Using the langserver with VSCode requires installing the Magic Racket extension.


jeapostrophe:
Thanks!

PR #42, opened 2021-04-04
Provide textDocument/rename and prepareRename

Runi-c:
Effective and expected renaming support. Until more workspace support is added, `prepareRename` rejects any rename requests for symbols not defined in the current file. Leads to more expected behavior than DrRacket in some cases.

https://user-images.githubusercontent.com/5150427/113500185-44527200-94d9-11eb-8257-1e615d3f731a.mp4

Also includes a few other improvements and fixes, such as a fix to signature & autocompletion providing while writing new code and some improvements to trigger characters for certain methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,16 +42,15 @@ You may need to restart your LSP runtime or your editor for `racket-langserver` 
   - *Note:* Currently only considers references within the current file.
 - **Document Highlight** (textDocument/documentHighlight)
 - **Diagnostics** (textDocument/publishDiagnostics)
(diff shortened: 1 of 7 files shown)


jackfirth:
Does this work with things that use `'sub-range-binders`, like struct fields?

Runi-c:
I didn't even think to try, but it does! It uses the binding information provided by DrRacket's arrow system, so anything supported by that should work as expected with renaming.


https://user-images.githubusercontent.com/5150427/113500656-fa6b8b00-94dc-11eb-88f2-4ea8a9765a40.mp4

jeapostrophe:
Awesome!

PR #44, opened 2021-04-05
Multithread syntax checks

Runi-c:
Vastly improve responsiveness by performing syntax checks on a separate thread, allowing on-type formatting, signature help, and completions to be provided much faster.

Sample on a large file (notice the error lines lagging behind, but other langserver features are very responsive):

https://user-images.githubusercontent.com/5150427/113528280-5ede2700-957d-11eb-9fd5-2f7eb8f4359d.mp4

PR #46, opened 2021-04-07
Fix messages getting mangled from threads

Runi-c:
There was a bug with the introduction of #44 causing the language server to sometimes mix together multiple JSON-RPC responses together, which causes that and all future responses to fail. By forcing all requests to go through one thread, this fix ensures the message chunks are always sent in intended order.


jeapostrophe:
Good catch

PR #51, opened 2021-04-12
Improve #lang errors, use default indenter if language has none

### LINKED ISSUE #49: Missing or invalid #lang line error for custom #lang
When I open files written in `#lang resyntax/testing/refactoring-test` ([example](https://github.com/jackfirth/resyntax/blob/master/default-recommendations/conditional-shortcuts-test.rkt)) I get the error "Missing or invalid #lang line" pointing at the `#lang` line. Based on [this code](https://github.com/jeapostrophe/racket-langserver/blob/b13132e284af977c3a7f6375e76e05f9134a796b/check-syntax.rkt#L101) it seems to be because my `#lang` implementation doesn't define an indenter. Yet DrRacket indents it using the default racket indenter just fine. Is this a bug in the langserver?

Runi-c:
Fixes #49

### DIFF
--- a/check-syntax.rkt
+++ b/check-syntax.rkt
@@ -66,15 +66,18 @@
                           #:source "Typed Racket"
                           #:message msg))))))
 
+(define default-indenter ((read-language (open-input-string "#lang racket/base")) 'drracket:indentation #f))
+
 (define (get-indenter doc-text)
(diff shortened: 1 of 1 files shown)


jeapostrophe:
Thanks @Runi-c !

PR #53, opened 2021-04-16
properly handle lexers with modes

Runi-c:
This fixes https://github.com/Eugleo/magic-racket/issues/43 and fixes https://github.com/Eugleo/magic-racket/issues/24
Some #langs return "stateful" lexers as a `(cons lexer mode)`, these didn't have any condition or proper handling.


jeapostrophe:
Excellent

PR #57, opened 2021-04-22
fix indenting when indenter returns #f

### LINKED ISSUE #56: at-exp meta language breaks indentation
With a lang-line like `#lang at-exp racket`, `read-language` returns the lang-info of `at-exp` only, which defers to Scribble's indenter. It seems like there's absolutely no built-in support for querying a meta-language like `at-exp` for its "base" language, so I'm not entirely sure what the correct method is for determining which indenter is the correct one to use on a particular line.

Some preliminary ideas:

- Just use the racket indenter when at-exp is used
- Dig into DrRacket source more and figure out how they handle this (hasn't been successful so far)
- Bug someone to properly implement `at-exp`'s indenter... it's probably not correct for it to just fully defer to the scribble indenter
- Manually re-implement the indenter... not sure how to determine if a given line is inside or outside of an @-exp

Runi-c:
Fixes #56 

Also removes the "default indenter" thing because I didn't realize that was fully already handled in the `indent-line!` helper and it wasn't working anyway.

### DIFF
--- a/check-syntax.rkt
+++ b/check-syntax.rkt
@@ -66,8 +66,6 @@
                           #:source "Typed Racket"
                           #:message msg))))))
 
-(define default-indenter ((read-language (open-input-string "#lang racket/base")) 'drracket:indentation #f))
-
 (define (get-indenter doc-text)
   (define lang-info 
     (with-handlers ([exn:fail:read? (lambda (e) 'missing)]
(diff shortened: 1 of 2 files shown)


jeapostrophe:
Thanks @Runi-c you're a star!

PR #63, opened 2021-07-30
fix indenter

dannypsnl:
originally, we get desired spaces and just applied, but in the case like
the following code:

```rkt
  (define x 1)

  (define y 2)
```

we end up get indent

```rkt
  (define x 1)
  
  (define y 2)
```

which is because we didn't check a line between code might be empty, and
them don't need indent


jeapostrophe:
Thanks!

PR #65, opened 2021-07-30
fix remove trailing whitespace

### LINKED ISSUE #64: remove trailing whitespace won't work for code after `#\"` until next string
I didn't check this condition, not just string will provide `"` character, `#\"` also. To fix this, need to check where `"` occurs is not provided from `#\"`.

dannypsnl:
check the char before `#\"` is not `#\\` shall be enough

resolve #64


jeapostrophe:
+1

PR #80, opened 2022-09-09
wrap identifier with `|` to get valid renaming

dannypsnl:
```racket
(define foo 1)
foo
```

If rename `foo` to `foo bar`, the new text should be `|foo bar|`, for convenience.

### DIFF
--- a/text-document.rkt
+++ b/text-document.rkt
@@ -342,17 +342,15 @@
      (define result
        (match decl
          [(Decl req? id left right)
-          (define new-text
-            (if (string-contains? new-name " ")
-                (string-append "|" new-name "|")
-                new-name))
           (cond [req? (json-null)]
                 [else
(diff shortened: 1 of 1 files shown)


text-document.rkt
```diff
@@ -342,13 +342,17 @@
      (define result
        (match decl
          [(Decl req? id left right)
+          (define new-text
+            (if (string-contains? new-name " ")
```
sorawee, line 346: That seems oddly too specific. What about other characters that need escaping as well?

Can `new-name` already contain `|`? If so, will `(string-append "|" new-name "|")` add redundant bars?

dannypsnl: Yes, I didn't think that carefully, I guess use `string->symbol` is enough.

rfindler:
If one is in a different #lang than #lang racket, is adding `|`s what you want? (Is it what you get?)

dannypsnl:
Can syncheck-annotation know the language?

PR #81, opened 2022-10-19
cleanup range-formatting

dannypsnl:


text-document.rkt
```diff
@@ -476,13 +476,12 @@
...
-                   (cond
-                     [(and edit edit2) (append (list edit edit2) (loop (add1 line)))]
-                     [edit (cons edit (loop (add1 line)))]
-                     [edit2 (cons edit2 (loop (add1 line)))]
-                     [else (loop (add1 line))]))))))
+                 (append (filter-map
+                          identity
+                          ; NOTE: The order is important somehow
+                          (list (remove-trailing-space! mut-doc-text skip-this-line? line)
+                                (indent-line! mut-doc-text indenter line)))
```
jeapostrophe, line 483: You reversed the order of the result, but preserved the order of the effect. In the original code `rts` is called first and `il` is called second, but `il`'s result is included first and `rts`'s second

jeapostrophe (changes requested):

text-document.rkt
dannypsnl: `remove-trailing-space!` and `indent-line!` both modify the `mut-doc-text`, thus, which one executes first is important.

But the returned `Edit[]` for the client side, will not be affected by the order, they will be done like in the same time.

jeapostrophe: Ok

jeapostrophe (approved):

PR #84, opened 2022-10-26
[method] `textDocument/inlayHint`

dannypsnl:
The current implementation will show "defined in file-xxx" for every definition. This is not quite helpful of course, to get more useful functionality, `check-syntax` should be able to collect the form of procedure like `(hash-ref hash key)`.

Then collector can collect this information and provide hints like `(hash-ref hash (make-hash) key 'a)`, notice that `hash` and `key` previous shows are inlay hints.

Below is the current situation:
圖片

### DIFF
--- a/doc-trace.rkt
+++ b/doc-trace.rkt
@@ -20,6 +20,7 @@
     (define docs (make-interval-map))
     (define completions (list))
     (define requires (make-interval-map))
+    (define definitions (make-hash))
     ;; decl -> (set pos ...)
     (define sym-decls (make-interval-map))
     ;; pos -> decl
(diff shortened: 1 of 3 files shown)


dannypsnl:
@jeapostrophe I let inlay-hint return an empty hint list for now since collecting metadata for structure or procedure from `syncheck-annotations<%>` is hard due to no direct supporting, but the method implementation is workable on the other hand, so I guess you can merge this for now and waiting for the future we know more about how to collect data and improve the editing experience.

jeapostrophe (approved):

jeapostrophe:
Excellent!

Issue #82, opened 2022-10-22
empty line added when `format()`


haoyun:
I am not sure if this is an issue on the language server side or on the neovim side (https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/racket_langserver.lua).

I am using this language server with neovim. Everytime the code is formatted with `:lua vim.lsp.buf.format()` via lsp, an empty line is added to the end of the file. So if autoformat is enable on save, lots of empty lines will be appended. 

I tested it does not happen in VSCode.

I've no idea how to inspect/debug the problem.

dannypsnl:
I’m looking into this problem since I have same stack, the weird thing is others language server has no this issue, only racket-langserver.

haoyun:
It's good to know that you can reproduce the bug. So, it's not a configuration problem on my side.

beelzebielsk:
I'm having the same problem, too.

6cdh:
I think it's fixed now.

PR #90, opened 2022-12-09
fix extra empty lines (#82) in formatting

6cdh:
For some reason, `end-line` is actually the number of lines of the file (0 based).

For example, the file have 3 lines. We finally have this [`TextEdit`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit) object in the response:

```json
{
  "newText": "",
  "range": {
    "start": {
      "character": 0,
      "line": 3
    },
    "end": {
      "character": 0,
      "line": 3
    }
  }
}
```

According to the protocol, this range is empty. It will depend on how the client process it.
I think it's a neovim bug. Anyway, we can avoid this useless response in our server.

### DIFF
--- a/text-document.rkt
+++ b/text-document.rkt
@@ -508,7 +508,7 @@
                (when (and (char=? #\" (send mut-doc-text get-character i))
                           (not (char=? #\\ (send mut-doc-text get-character (sub1 i)))))
                  (set! skip-this-line? (not skip-this-line?))))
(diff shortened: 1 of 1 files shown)


dannypsnl:
fantastic

jeapostrophe (approved):

jeapostrophe:
Thank you!

PR #93, opened 2023-01-28
clean msg-io.rkt

6cdh:
This PR did these changes:

* Remove unused variables `verbose-io?` because it's never changed.
* don't export `display-message` because it's not safe when writing to the output port. It should be an internal function. It's also never used outside the module.
* rewrite the `display-message` function to make it clear and slightly faster.
* fix a bug that the specified output-port of `display-message/flush` isn't passed to `display-message` correctly. It makes `test-main` runnable again.

### DIFF
--- a/msg-io.rkt
+++ b/msg-io.rkt
@@ -2,11 +2,9 @@
 (require json
          racket/contract/base
          racket/match
-         racket/string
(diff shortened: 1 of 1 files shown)


jeapostrophe (approved):

jeapostrophe:
Thanks!

PR #96, opened 2023-02-09
add tests and fix code actions

6cdh:
I added a lot of tests, and fixed a bug that code action don't actually work.

### DIFF
--- a/tests/client.rkt
+++ b/tests/client.rkt
@@ -1,14 +1,16 @@
-#lang racket/base
+#lang racket
 
 ;; This module provides a simple incomplete lsp client for test.
 
 (provide with-racket-lsp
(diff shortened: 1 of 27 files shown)


tests/client.rkt
```diff
@@ -1,14 +1,16 @@
-#lang racket/base
+#lang racket
```
jeapostrophe, line 1: `racket/base` is very desirable for libraries. Since this is a test, it isn't a big deal, but FYI

jeapostrophe (approved):

jeapostrophe:
Excellent!

tests/client.rkt
6cdh: ok

PR #100, opened 2023-03-25
Fix indentation after new line at first line of expression

jryans:
Adding a new line at the end of the first line of an existing expression would force you back to the start of the line unexpectedly.

For example, with the following:

```racket
(define (bob); <- place cursor here, then press Return
  (+ 1 2))
```

Previously the langserver left the new line unindented. This fixes the indentation path to instead indent the new line as expected.

### DIFF
--- a/tests/client.rkt
+++ b/tests/client.rkt
@@ -7,7 +7,7 @@
          client-send
          client-wait-response
          make-request
-         make-response
+         make-expected-response
          make-notification)
 
 (require racket/os
@@ -80,7 +80,7 @@
   (cond [(not params) req]
         [else (hash-set req 'params params)]))
 
-(define/contract (make-response request result)
+(define/contract (make-expected-response request result)
   (-> jsexpr? jsexpr? jsexpr?)
 
   (define res
--- a/tests/client.rkt
+++ b/tests/client.rkt
@@ -7,6 +7,7 @@
          client-send
          client-wait-response
          make-request
+         make-response
          make-notification)
 
 (require racket/os
@@ -78,6 +79,15 @@
             'method method))
   (cond [(not params) req]
         [else (hash-set req 'params params)]))
+
+(define/contract (make-response request result)
+  (-> jsexpr? jsexpr? jsexpr?)
+
+  (define res
+    (hasheq 'jsonrpc "2.0"
+            'id (hash-ref request 'id)))
+  (cond [(not result) res]
+        [else (hash-set res 'result result)]))
 
 (define/contract (make-notification method params)
   (-> string? jsexpr? jsexpr?)
--- a/text-document.rkt
+++ b/text-document.rkt
@@ -508,7 +508,7 @@
      (define end-pos (send doc-text last-position))
      (define start (abs-pos->Pos doc-text 0))
      (define end (abs-pos->Pos doc-text end-pos))
-     (success-response id (format! uri start end 'document))]
+     (success-response id (format! uri start end))]
     [_
      (error-response id INVALID-PARAMS "textDocument/formatting failed")]))
 
@@ -518,7 +518,7 @@
     ;; XXX We're ignoring 'options' for now
     [(hash-table ['textDocument (DocIdentifier #:uri uri)]
                  ['range (Range  #:start start #:end end)])
-     (success-response id (format! uri start end 'range))]
+     (success-response id (format! uri start end))]
     [_
      (error-response id INVALID-PARAMS "textDocument/rangeFormatting failed")]))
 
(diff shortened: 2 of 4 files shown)


jryans:
@6cdh, @dannypsnl: My current approach here effectively reverts https://github.com/jeapostrophe/racket-langserver/pull/63 and https://github.com/jeapostrophe/racket-langserver/pull/92... Please take a look at this change. Is there a way we can adjust it to preserve the behaviour you were each trying to fix in those PRs?

6cdh:
I think the root of the issue is `on-type-formatting` calls `range-formatting`.

Based on my experience,
* when `on-type-formatting` is called, the programmer is typing, and we expect it has indentation for an empty line.
* when `range-formatting` or `formatting` is called, the programmer completed editing, and we expect it remove trailing whitespaces.

These two cases are in conflict.
To solve this issue, we could have two versions of `range-formatting`, one for editing mode and one for not.

dannypsnl:
Just check https://github.com/jeapostrophe/racket-langserver/blob/master/text-document.rkt#L582 said it reuse `range-formatting!`, so just as @6cdh suggestion, you might like to write another implementation here

jryans:
Ah okay, thanks for the suggestions! 🙂 I could potentially pass down a formatting type symbol (on-type, range, document) into `range-formatting!` so we can tweak behaviour based on which one is being called.

I think I'll also investigate the details of how DrRacket and also other language extensions typically handle these different formatting types.

jryans:
Okay, this now preserves the tweaks from #63 and #92 while still allowing empty line indentation while typing.

@6cdh and @dannypsnl, if you could take another look to see if I've missed anything, that would be great. (It might be easier to review commit by commit.)

text-document.rkt
```diff
@@ -567,24 +532,57 @@
...
+                                      0))
+           (abs-pos->Pos doc-text pos))]))
+     (success-response id (format! uri start end 'on-type))]
     [_
-     (error-response id INVALID-PARAMS "textDocument/onTypeformatting failed")]))
+     (error-response id INVALID-PARAMS "textDocument/onTypeFormatting failed")]))
+
+;; Shared path for all formatting requests
+;; `kind` allows diverging in a few places while sharing the core operation
+(define (format! uri start end kind)
```
dannypsnl, line 552: we might need a contract to ensure `kind` is valid?

text-document.rkt
```diff
@@ -567,24 +532,57 @@
...
+          (when (and (char=? #\" (send mut-doc-text get-character i))
+                      (not (char=? #\\ (send mut-doc-text get-character (sub1 i)))))
+            (set! skip-this-line? (not skip-this-line?))))
+        (if (> line end-line)
+            null
+            (append (filter-map
+                      identity
+                      ; NOTE: The order is important somehow
+                      (list (remove-trailing-space! mut-doc-text skip-this-line? line)
+                            (indent-line! mut-doc-text indenter line kind)))
```
dannypsnl, line 584: The order is important, since remove space is "remove" the position. Consider this case:
```
xxx•••
indent to ~>
  xxx•••
then remove space cut off position ~>
  x••
```

text-document.rkt
```diff
@@ -619,7 +618,7 @@
   (cond
     [(not (number? desired-spaces)) #f]
     [(= current-spaces desired-spaces) #f]
-    [(= line-length 0) #f]
+    [(and (not (eq? kind 'on-type)) (= line-length 0)) #f]
```
dannypsnl, line 621: If we only need `'on-type`, maybe we can make it a flag for now? e.g.
```racket
(define (xxx #:on-type? [on-type? #f])
  ...)
```

text-document.rkt
jryans: Right, it makes sense to me... The comment above is just your own comment from https://github.com/jeapostrophe/racket-langserver/commit/ad7c0c958968d526c1b27c4d40959d737fd81111 that I copied while extracting a shared path. 😄

Should we change the comment to explain the details or perhaps remove it?

text-document.rkt
jryans: Potentially, though it doesn't seem like any of the "internal" functions in this file use contracts at the moment... Maybe it's not needed if we switch to a flag like you suggested below?

text-document.rkt
jryans: Ah sure, I'm happy to use a flag instead for this case. I'll wait a bit to see if there are any other comments, and then I'll make this change.

text-document.rkt
dannypsnl: Oh! Did notice that from myself, but I think this is a good chance to add this into comment.

tests/client.rkt
```diff
@@ -79,11 +80,20 @@
   (cond [(not params) req]
         [else (hash-set req 'params params)]))
 
+(define/contract (make-response request result)
```
6cdh, line 83: Perhaps we could consider renaming it to `make-expected-response` as `make-response` is not the responsibility of the client.

text-document.rkt
6cdh: I agree with dannypsnl's comment.

text-document.rkt
jryans: Sounds good, I've added this detail in a longer comment.

jryans:
All review comments have been addressed, please take another look. 😄

6cdh (approved):

dannypsnl (approved):

jeapostrophe (approved):

jeapostrophe:
Thank you so much everyone!

Issue #102, opened 2023-04-03
A `#;` without something to comment out yields an unrecoverable error


Release-Candidate:
Entering a #; in the last line of a file results in an unrecoverable error, deleting the `#;` does not make the error go away.
It works when first entering the S-expression to comment out and then adding the `#;`.

like:

```racket
#lang racket

#;
```

Output of the LSP:

```
sub1: contract violation
  expected: number?
  given: #f
  context...:
   /Users/roland/Library/Racket/8.8/pkgs/racket-langserver/check-syntax.rkt:19:0
   /Applications/Racket v8.8/collects/racket/private/more-scheme.rkt:163:2: select-handler/no-breaks
   /Applications/Racket v8.8/collects/racket/logging.rkt:43:0: with-intercepted-logging
   /Users/roland/Library/Racket/8.8/pkgs/racket-langserver/check-syntax.rkt:77:0: check-syntax
   /Applications/Racket v8.8/collects/racket/contract/private/arrow-val-first.rkt:486:18
   /Users/roland/Library/Racket/8.8/pkgs/racket-langserver/queue-only-latest.rkt:9:0: do-check-and-stuff!
   /Users/roland/Library/Racket/8.8/pkgs/racket-langserver/queue-only-latest.rkt:33:0: check
```
See: https://github.com/Eugleo/magic-racket/issues/104

PR #104, opened 2023-04-07
Use end-of-file as fallback position for exceptions

jryans:
Some reader exceptions do not have position information. This uses end-of-file as a reasonable guess, so we can at least highlight some part of the file. Handing this case resolves a langserver contract error.

Fixes https://github.com/jeapostrophe/racket-langserver/issues/102


jeapostrophe (approved):
Great!

Issue #101, opened 2023-04-02
LSP does not like anything 'above' `#lang`


Release-Candidate:
Anything before `#lang` in a file causes an exception. It does not matter whether there is a comment, an empty line,... -  

Example - adding an empty line (which isn't visible in the rendered Markdown) above `#lang`:
```racket

#lang racket
```

Exception:
```
Caught exn in request "textDocument/documentSymbol"
rest: contract violation
  expected: (and/c list? (not/c empty?))
  given: #<void>
  context...:
   /Users/roland/Library/Racket/8.8/pkgs/racket-langserver/text-document.rkt:232:0: get-symbols
   /Users/roland/Library/Racket/8.8/pkgs/racket-langserver/text-document.rkt:463:0: document-symbol
   /Applications/Racket v8.8/collects/racket/contract/private/arrow-val-first.rkt:486:18
   /Users/roland/Library/Racket/8.8/pkgs/racket-langserver/methods.rkt:26:0: process-message
   /Applications/Racket v8.8/collects/racket/contract/private/arrow-val-first.rkt:486:18
   /Users/roland/Library/Racket/8.8/pkgs/racket-langserver/main.rkt:60:2: consume

[Error - 21:37:57] Request textDocument/documentSymbol failed.
  Message: internal error in method "textDocument/documentSymbol"
  Code: -32603 
```

dannypsnl:
This should be fixed, @Release-Candidate would you like to check it?

Release-Candidate:
Yes, thanks. This works for me with the latest LSP version.

Issue #97, opened 2023-03-14
‘hash-ref: no value found for key’ for certain files


bluebear94:
Steps to reproduce (using Magic Racket extension for VSCode):

* Clone the [official Pollen repository](https://git.matthewbutterick.com/mbutterick/pollen)
* Open `pollen/render.rkt`
* Try to hover over some text

Expected behavior: hovering shows a tooltip for the text (if applicable)

Actual behavior: the language server gives the following errors:

```
Caught exn:
standard-module-name-resolver: contract violation
  expected: module-path?
  given: '(just-space #f (rename scribble/manual racketmodlink racketmodlink))
  context...:
   /home/felirovas/.local/share/racket/8.7/pkgs/racket-langserver/autocomplete.rkt:106:0: walk-module
   /home/felirovas/.local/share/racket/8.7/pkgs/racket-langserver/check-syntax.rkt:111:10
   /usr/share/racket/collects/racket/logging.rkt:43:0: with-intercepted-logging
   /home/felirovas/.local/share/racket/8.7/pkgs/racket-langserver/check-syntax.rkt:77:0: check-syntax
   /usr/share/racket/collects/racket/contract/private/arrow-val-first.rkt:486:18
   /home/felirovas/.local/share/racket/8.7/pkgs/racket-langserver/text-document.rkt:68:0: did-open!
   /usr/share/racket/collects/racket/contract/private/arrow-val-first.rkt:486:18
   [repeats 1 more time]
   /home/felirovas/.local/share/racket/8.7/pkgs/racket-langserver/main.rkt:60:2: consume

Caught exn in request "textDocument/documentSymbol"
hash-ref: no value found for key
  key: 'file:///home/felirovas/conlang/ncv9/pollen/pollen/render.rkt
  context...:
   /home/felirovas/.local/share/racket/8.7/pkgs/racket-langserver/text-document.rkt:463:0: document-symbol
   /usr/share/racket/collects/racket/contract/private/arrow-val-first.rkt:486:18
   /home/felirovas/.local/share/racket/8.7/pkgs/racket-langserver/methods.rkt:26:0: process-message
   /usr/share/racket/collects/racket/contract/private/arrow-val-first.rkt:486:18
   /home/felirovas/.local/share/racket/8.7/pkgs/racket-langserver/main.rkt:60:2: consume

[Error - 3:09:20 AM] Request textDocument/documentSymbol failed.
  Message: internal error in method "textDocument/documentSymbol"
[...]

PR #110, opened 2023-07-08
Add document check job queue, abstract out lsp-editor%

### LINKED ISSUE #109: 100% CPU and never terminate for macro with infinite recursion
When I open a file with the following code in vscode:

```racket
#lang racket

(define-syntax (f stx)
  stx)

f
```

It reachs 100% CPU and doesn't terminate. I've been running it for a few minutes and it's still running.

I had this problem for a while, but just found a minimal sample to reproduce it.

6cdh:
I did
* remove all `match` on struct `Doc`, Use struct accessor instead. So `doc` can contain extra hidden fields.
* add a extra middle layer `doc.rkt` between check-syntax, doc-trace and text-document, and move some code from `text-document.rkt` to `doc.rkt`. The `Doc-text`, `Doc-trace` should not be called in `text-document.rkt`, it should call a proxy method in `doc.rkt` instead.
* isolate `check-syntax` so that it's not called here and there. It's only called in `doc-run-check-syntax` now.
* clear `check-syntax`. It no longer sends diagnostics directly, instead, it returns diagnostics.
* use a unified editor class `lsp-editor%` instead of using `text%` or `racket:text%` everywhere. It would be useful if we write our own implementation to solve the gui problem.
* add a 90 seconds timeout for `expand`, should fix #109.
* improve the `queue-only-latest.rkt` implementation.
  * in the old design, there is at most a running `check-syntax` job and a pending job in the whole program. Any new incoming job will replace the pending job.
  * in the new design in `scheduler.rkt`, every document that is identified by uri can have a running job. Any new incoming job for this document will kill the old running job and run itself.
* make the initial trace object to be an empty `build-trace%` object not `#f`. It will fix many crash problem if `check-syntax` never succeed.
* other small improvements

The following issues will be fixed
* #101
* #97


check-syntax.rkt
```diff
@@ -21,6 +20,13 @@
   (cond
     ;; typed racket support: don't report error summaries
     [(string-prefix? msg "Type Checker: Summary") (list)]
+    [(exn:fail:resource? exn)
+     (list (Diagnostic #:range (Range #:start (Pos #:line 0 #:char 0)
+                                      #:end (Pos #:line 0 #:char 0))
+                       #:severity Diag-Hint
+                       #:source "Expander"
+                       #:message "the expand time has exceeded the 90s limit.\
+                        Check if your macro is infinitely recursive."))]
```
dannypsnl, line 29: "is infinitely expanding"?

doc.rkt
```diff
@@ -0,0 +1,274 @@
...
+;; the only place where really run check-syntax
+(define (doc-run-check-syntax doc)
+  (define (task)
+    (set-Doc-checked?! doc #f)
+    (match-define (list new-trace diags)
+      (check-syntax (uri->path (Doc-uri doc)) (Doc-text doc) (Doc-trace doc)))
+    (send-diagnostics doc diags)
+    (set-Doc-trace! doc new-trace)
+
+    (set-Doc-checked?! doc #t))
```
dannypsnl, line 41: Could this function be called in concurrency? If so, then the `checked?` modification is not safe, need a lock here

doc.rkt
```diff
@@ -0,0 +1,274 @@
...
+(define (doc-line-start-pos doc line)
+  (send (Doc-text doc) line-start-pos line))
+
+(define (doc-line-end-pos doc line)
+  (send (Doc-text doc) line-end-pos line))
+
+(define (doc-endpos doc)
+  (send (Doc-text doc) end-pos))
+
+(define (doc-find-containing-paren doc pos)
```
dannypsnl, line 109: What's the purpose of this function? Checking paren balancing and stop invalid file?

scheduler.rkt
```diff
@@ -0,0 +1,29 @@
...
+;; no matter if the old one is running or completed
+(define (schedule)
+  (define open-doc (make-hash))
+  (let loop ()
+    (sync (handle-evt incoming-jobs-ch
+                      (λ (data)
+                        (match-define (cons path task) data)
+                        (when (hash-has-key? open-doc path)
+                          (define th (hash-ref open-doc path))
+                          (when (not (thread-dead? th))
```
dannypsnl, line 18: Maybe use `unless (thread-dead? th)`?

text-document.rkt
```diff
@@ -118,12 +121,14 @@
...
-     (unless (uri-is-path? uri)
-       (error 'hover "uri is not a path"))
-     (match-define (doc doc-text doc-trace)
-       (hash-ref open-docs (string->symbol uri)))
+     #:when (uri-is-path? uri)
+
+     (define this-doc (hash-ref open-docs (string->symbol uri)))
+
+     (define doc-trace (Doc-trace this-doc))
+
```
dannypsnl, line 129: Misc. I think we don't need these blank line?

doc.rkt
6cdh: No. There's only one `task` running for a given document at anytime.
The `set-Doc-*` methods are expected to be safe.

doc.rkt
6cdh: It's a new wrapper for the existing `find-containing-paren`. The function finds the closest pair of brackets the containing `pos` If I understand it correctly.

dannypsnl (approved):

dannypsnl:
I think the title can be "Add document check job queue, abstract out `lsp-editor%`"

editor.rkt
```diff
@@ -9,16 +9,32 @@
...
 
     ;; replace text at (range start end) with str
     (define/public (replace str start end)
-      (send core insert str start end))
+      (with-handlers ([exn?
+                       (λ _
+                         (reload!)
+                         ;; only retry once
+                         (send core insert str start end))])
+        (send core insert str start end)))
```
dannypsnl, line 37: `send core replace ...`?

dannypsnl (changes requested):

6cdh:
The new `fix editor crash` commit fixed a random editor crash with this message:
```
sequence-contract-violation: negative: method insert cannot be called
```

It's a well known bug that no one knows how it happened. https://github.com/racket/drracket/issues/522
I found it today and added a workaround that recreate a new `text%` to replace to old one.

6cdh:
Everything should work fine right now to be merged.

editor.rkt
6cdh: There is no such a method for `text%`.

dannypsnl: Oh, I get the difference.

dannypsnl (approved):

jeapostrophe (approved):

PR #116, opened 2023-07-21
Refine package dependency

shhyou:
- Add `sandbox-lib` to incorporate the use of `with-limits`
- Depend only on `chk-lib` since there's no need to pull `chk-doc`

I had wished that `drracket-tool-lib` could be replaced by `drracket-tool-text-lib`,
but that would break installation on earlier (< 8.4) versions of Racket.

### DIFF
--- a/info.rkt
+++ b/info.rkt
@@ -3,13 +3,17 @@
 (define deps '("base"
                "compatibility-lib"
                "data-lib"
+               ;; Newer packages need only to depend on drracket-tool-text-lib.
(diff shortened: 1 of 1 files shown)


dannypsnl (approved):

dannypsnl:
@shhyou should I close this PR? I see the newer one include this

shhyou:
I prefer to keep them separate and rebase the other one after this is merged

6cdh (approved):

PR #119, opened 2023-07-25
Respect formatting options

### LINKED ISSUE #112: support formatting options
Link: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#formattingOptions

```ts
interface FormattingOptions {
	/**
	 * Size of a tab in spaces.
	 */
	tabSize: uinteger

	/**
	 * Prefer spaces over tabs.
	 */
	insertSpaces: boolean;

	/**
	 * Trim trailing whitespace on a line.
	 *
	 * @since 3.15.0
	 */
	trimTrailingWhitespace?: boolean;

	/**
	 * Insert a newline character at the end of the file if one does not exist.
	 *
	 * @since 3.15.0
	 */
	insertFinalNewline?: boolean;

	/**
	 * Trim all newlines after the final newline at the end of the file.
	 *
	 * @since 3.15.0
	 */
	trimFinalNewlines?: boolean;

	/**
	 * Signature for further properties.
	 */
	[key: string]: boolean | integer | string;
}
```

6cdh:
I also rewrote some formatting helper functions. It should fix #112

### DIFF
--- a/doc.rkt
+++ b/doc.rkt
@@ -227,6 +227,9 @@
   ;; For an old lsp-editor% that already contains some text,
   ;; it would not work.
   (define mut-doc-text (send doc-text copy))
+  ;; replace \t with spaces at line `(sub1 start-line)`
+  ;; as we cannot make `compute-racket-amount-to-indent`
+  ;; to respect the given tab size
   (replace-tab! mut-doc-text
                 (max 0 (sub1 start-line))
                 (FormattingOptions-tabSize fo))
--- a/doc.rkt
+++ b/doc.rkt
@@ -211,7 +211,7 @@
 ;; Shared path for all formatting requests
 (define (format! this-doc st-ln st-ch ed-ln ed-ch
                  #:on-type? [on-type? #f]
-                 #:formatting-options fo)
+                 #:formatting-options opts)
   (define doc-text (Doc-text this-doc))
   (define doc-trace (Doc-trace this-doc))
 
@@ -228,7 +228,7 @@
   ;; to respect the given tab size
   (replace-tab! mut-doc-text
                 (max 0 (sub1 start-line))
-                (FormattingOptions-tabSize fo))
+                (FormattingOptions-tabSize opts))
 
   (define indenter-wp (indenter-wrapper indenter mut-doc-text on-type?))
   (define skip-this-line? #f)
@@ -250,7 +250,7 @@
                      ;; position. If we were to instead call `indent-line!` first and then
                      ;; `remove-trailing-space!` second, the remove step could result in
                      ;; losing user entered code.
-                     (list (if (false? (FormattingOptions-trimTrailingWhitespace fo))
+                     (list (if (false? (FormattingOptions-trimTrailingWhitespace opts))
                                #f
                                (remove-trailing-space! mut-doc-text skip-this-line? line))
                            (indent-line! mut-doc-text indenter-wp line)))
--- a/doc.rkt
+++ b/doc.rkt
@@ -266,7 +266,9 @@
 
 (define (indenter-wrapper indenter doc-text on-type?)
   (λ (line)
-    (cond [(and (not on-type?) (= 0 (send doc-text line-end-pos line)))
+    (cond [(and (not on-type?)
+                (= 0 (- (send doc-text line-start-pos line)
+                        (send doc-text line-end-pos line))))
            #f]
           [else
            (define line-start (send doc-text line-start-pos line))
--- a/doc.rkt
+++ b/doc.rkt
@@ -267,8 +267,8 @@
 (define (indenter-wrapper indenter doc-text on-type?)
   (λ (line)
     (cond [(and (not on-type?)
-                (= 0 (- (send doc-text line-start-pos line)
-                        (send doc-text line-end-pos line))))
+                (= (send doc-text line-start-pos line)
+                   (send doc-text line-end-pos line)))
            #f]
           [else
            (define line-start (send doc-text line-start-pos line))
--- a/doc.rkt
+++ b/doc.rkt
@@ -208,7 +208,9 @@
 ;; formatting ;;
 
 ;; Shared path for all formatting requests
-(define (format! this-doc st-ln st-ch ed-ln ed-ch #:on-type? [on-type? #f])
+(define (format! this-doc st-ln st-ch ed-ln ed-ch
+                 #:on-type? [on-type? #f]
+                 #:formatting-options fo)
   (define doc-text (Doc-text this-doc))
   (define doc-trace (Doc-trace this-doc))
 
--- a/doc.rkt
+++ b/doc.rkt
@@ -222,10 +222,6 @@
   (define start-line (send doc-text at-line start-pos))
   (define end-line (send doc-text at-line end-pos))
 
-  ;; copy a lsp-editor% with a new specified tabsize
-  ;; We can only specify the tab size for a new empty lsp-editor%.
-  ;; For an old lsp-editor% that already contains some text,
-  ;; it would not work.
   (define mut-doc-text (send doc-text copy))
   ;; replace \t with spaces at line `(sub1 start-line)`
   ;; as we cannot make `compute-racket-amount-to-indent`
--- a/doc.rkt
+++ b/doc.rkt
@@ -14,6 +14,7 @@
          racket/set
          racket/list
          racket/string
+         racket/bool
          data/interval-map
          syntax-color/module-lexer
          syntax-color/racket-lexer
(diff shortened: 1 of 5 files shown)


text-document.rkt
```diff
@@ -469,9 +470,10 @@
     ;; XXX We're ignoring 'options' for now
     [(hash-table ['textDocument (DocIdentifier #:uri uri)]
                  ['range (Range  #:start (Pos #:line st-ln #:char st-ch)
-                                 #:end (Pos #:line ed-ln #:char ed-ch))])
+                                 #:end (Pos #:line ed-ln #:char ed-ch))]
+                 ['options (as-FormattingOptions fo)])
```
dannypsnl, line 474: The comment is outdated now

text-document.rkt
```diff
@@ -453,13 +453,14 @@
 (define (formatting! id params)
   (match params
     ;; We're ignoring 'options for now
-    [(hash-table ['textDocument (DocIdentifier #:uri uri)])
+    [(hash-table ['textDocument (DocIdentifier #:uri uri)]
+                 ['options (as-FormattingOptions fo)])
```
dannypsnl, line 457: here outdated, too

editor.rkt
```diff
@@ -76,7 +78,7 @@
 
     (define/public (copy)
       (define new-editor (new lsp-editor%))
-      (send new-editor set-core! (send core copy-self))
+      (send new-editor insert (send this get-text) 0)
```
dannypsnl, line 81: Could you explain this changes? Is that means all we need copy text to `new-editor`?

struct.rkt
```diff
@@ -13,6 +16,31 @@
...
+
+;; request error
+
+(define request-err-object (gensym 'request-error))
+
+(define (request-err? x)
+  (eq? x request-err-object))
+
+(provide request-err-object
+         request-err?)
```
dannypsnl, line 42: What we need `gensym` to protect it from outside?

If we really need this, I also recommend another implementation which is based on reference:
```racket
(struct request-err ())

(define request-err-object (request-err))
(define (request-err? e)
  (eq? request-err-object e))
```

struct.rkt
```diff
@@ -62,3 +90,53 @@
...
          make-Range
          Range->hash)
+
+(struct FormattingOptions
+  (tabSize
+   insertSpaces
+   trimTrailingWhitespace
+   insertFinalNewline
+   trimFinalNewlines
+   key))
```
dannypsnl, line 100: Not sure, but do we always use this style for structure fields? Or JSON requires that?

struct.rkt
```diff
@@ -62,3 +90,53 @@
...
+                            #:insertFinalNewline (hash-ref js 'insertFinalNewline undef-object)
+                            #:trimFinalNewlines (hash-ref js 'trimFinalNewlines undef-object)
+                            #:key (hash-ref js 'key undef-object))))
+
+(define-match-expander as-FormattingOptions
+  (lambda (stx)
+    (syntax-parse stx
+      [(_ name)
+       #'(and (? hash?)
+              (app jsexpr->FormattingOptions (and name (not (? request-err?)))))])))
```
dannypsnl, line 137: Add an example in comment for this pattern macro would be great

struct.rkt
```diff
@@ -62,3 +90,53 @@
...
+      FormattingOptions?)
+
+  (FormattingOptions tabSize
+                     insertSpaces
+                     trimTrailingWhitespace
+                     insertFinalNewline
+                     trimFinalNewlines
+                     key))
+
+(define (jsexpr->FormattingOptions js)
```
dannypsnl, line 123: `js` often stands for **JavaScript**, I think we better use abbreviation like `json` instead of compress it futher

doc.rkt
```diff
@@ -242,60 +250,68 @@
                      ;; position. If we were to instead call `indent-line!` first and then
                      ;; `remove-trailing-space!` second, the remove step could result in
                      ;; losing user entered code.
-                     (list (remove-trailing-space! mut-doc-text skip-this-line? line)
-                           (indent-line! mut-doc-text indenter line #:on-type? on-type?)))
+                     (list (if (false? (FormattingOptions-trimTrailingWhitespace fo))
+                               #f
+                               (remove-trailing-space! mut-doc-text skip-this-line? line))
```
dannypsnl, line 255: I think
```racket
(and (FormattingOptions-trimTrailingWhitespace fo)
     (remove-trailing-space! mut-doc-text skip-this-line? line))
```
will work

dannypsnl (changes requested):

dannypsnl:
Oh, and as a personal opinion, maybe all `fo` can be just replaced by `opts` in the program. Since they are obviously in context.

editor.rkt
6cdh: Yes. We only need to copy text to the new editor.

All of the public methods of `lsp-editor%` except the indent method are simply insert/delete/change/query text strings.
It does not require additional states.

The private states that `core` contains are always the default. So we just ignore them.

struct.rkt
6cdh: I think there is no much difference here.

My initial idea was to use structs.
Then I remembered `gensym` which allowed me to avoid define additional structs. It generates an uninterned symbol so it's safe to compare with it.

editor.rkt
dannypsnl: Ok, your explanation can be a comment here, then this part is ok

struct.rkt
6cdh: It's intended to be same as the names in LSP specification.

Maybe we can have a macro to automatically converts these names to racket name style.
Then we only use the LSP style names in `jsexpr->FormattingOptions`, and racket style names everywhere else.

struct.rkt
6cdh: ok

text-document.rkt
6cdh: ok

text-document.rkt
6cdh: ok

struct.rkt
dannypsnl: Oh, I make a measure, symbol comparison got less cpu and gc time. Just mark this resolved.

Appendix test code

```rkt
#lang racket

(define N 100000000)

(struct request-err-ty ())

(define request-err-object (request-err-ty))
(define (request-err? e)
  (eq? request-err-object e))

(time
 (for ([i N])
   (request-err? request-err-object)
   (request-err? 'abc)))

(define request-err-object2 (gensym 'request-error))

(define (request-err2? x)
  (eq? x request-err-object2))

(time
 (for ([i N])
   (request-err2? request-err-object2)
   (request-err2? 'abc)))
```

struct.rkt
dannypsnl: It would be great if that is possible

struct.rkt
6cdh: we can use name `jsexpr` instead, the same as the json library used.

doc.rkt
6cdh: No. `FormattingOptions-trimTrailingWhitespace` could be omited in the message and will be an `undef-object` which is not `#f`.

6cdh:
> Oh, and as a personal opinion, maybe all fo can be just replaced by opts in the program. Since they are obviously in context.

ok

struct.rkt
dannypsnl: ok

doc.rkt
dannypsnl: If we assign `#f` as it's default value?

struct.rkt
6cdh: What about `FormattingOptions` and other top level structs? Their name are also inconsistent with Racket style.

doc.rkt
dannypsnl: Ok, then do we have chance to make it be `#f` when it's undefined?

struct.rkt
dannypsnl: Same, but I think maybe let's have another PR for these changes.

doc.rkt
6cdh: No. It naturally has three different states: undefined, boolean true, boolean false.
A trick to merge two states just for simpler syntactic sugar expression instead of simpler logic feels not good.

struct.rkt
6cdh: ok. I'm going to rename these fields.

doc.rkt
dannypsnl: Ok, I get it. You can mark this as resolved now

dannypsnl (approved):

6cdh:
The automatic renaming is in progress though. There are some difficulties. It could be in another PR.

PR #132, opened 2024-09-07
fix: remove unnecessary `#lang` diagnostic in drracket-created files

### LINKED ISSUE #86: Server requires #lang line even though #reader is present.
By default, opening a file created in DrRacket in VSCode throws the error `Missing or invalid #lang line` from `define lang-diag` as seen in `check-syntax.rkt`. However this file compiles and runs just fine. Below is an example of the first three lines inserted by DrRacket.

```
;; The first three lines of this file were inserted by DrRacket. They record metadata
;; about the language level of this file in a form that our tools can easily process.
#reader(lib "htdp-intermediate-lambda-reader.ss" "lang")((modname 1-family) (read-case-sensitive #t) (teachpacks ()) (htdp-settings #(#t constructor repeating-decimal #f #t none #f () #f)))
```

virchau13:
Student language files created by DrRacket do not have `#lang` directives; they only have `#reader` directives instead. (See e.g. https://github.com/jeapostrophe/racket-langserver/issues/86). The "Missing `#lang`" diagnostic should not be reported in these files.

Fixes #86.

### DIFF
--- a/check-syntax.rkt
+++ b/check-syntax.rkt
@@ -117,14 +117,21 @@
                           #:message msg))))))
 
 (define (get-indenter doc-text)
+  (define text (send doc-text get-text))
   (define lang-info
     (with-handlers ([exn:fail:read? (lambda (e) 'missing)]
                     [exn:missing-module? (lambda (e) #f)])
-      (read-language (open-input-string (send doc-text get-text)) (lambda () 'missing))))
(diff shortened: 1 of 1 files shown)


dannypsnl (approved):

jeapostrophe:
Thank you!

PR #133, opened 2024-09-09
Add initial Semantic token

6cdh:
Supported tokens: `(variable function string number regexp)`
Supported modifiers: `(definition)`
Supported method: [textDocument/semanticTokens/range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_rangeRequest) and [textDocument/semanticTokens/full](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_fullRequest).

There are some difficulties to support tokens like comment, struct, class, etc.

I don't know how to implement a embeded sexp comment reader. And struct, class, etc requires a full functional type infer.

image:

Screenshot 2024-09-10 010323

image with a semantic highlight plugin which gives different variable different color:

Screenshot 2024-09-10 010355

### DIFF
--- a/doc.rkt
+++ b/doc.rkt
@@ -9,6 +9,7 @@
          "path-util.rkt"
          "doc-trace.rkt"
          "struct.rkt"
+         "highlight.rkt"
          racket/match
          racket/class
          racket/set
@@ -315,6 +316,40 @@
                                   #:end (Pos #:line line #:char really-indent))
                    #:newText new-text)]))
 
+(define (token-type-encoding token)
+  (index-of *semantic-token-types* (SemanticToken-type token)))
+
+(define (token-modifier-encoding token)
+  (define indexes (indexes-where *semantic-token-modifiers*
+                                 (λ (m) (memq m (SemanticToken-modifiers token)))))
+  (for/sum ([index indexes])
+    (expt 2 index)))
+
+(define (token-encoding doc token prev-pos)
+  (define-values (line ch) (doc-line/ch doc (SemanticToken-start token)))
+  (define-values (prev-line prev-ch) (doc-line/ch doc prev-pos))
+  (define delta-line (- line prev-line))
+  (define delta-start
+    (if (= line prev-line)
+        (- ch prev-ch)
+        ch))
+  (define len (- (SemanticToken-end token) (SemanticToken-start token)))
+  (define type (token-type-encoding token))
+  (define modifier (token-modifier-encoding token))
+  (values delta-line delta-start len type modifier))
+
+(define (doc-full-tokens doc path)
+  (define tokens (collect-semantic-tokens (Doc-text doc) (uri->path path)))
+  (for/fold ([result '()]
+             [prev-pos 0]
+             #:result (let ()
+                        (flatten (reverse result))))
+            ([token tokens])
+    (define-values (delta-line delta-start len type modifier)
+      (token-encoding doc token prev-pos))
+    (values (cons (list delta-line delta-start len type modifier) result)
+            (SemanticToken-start token))))
+
 (provide Doc-trace
          new-doc
          doc-checked?
@@ -329,4 +364,6 @@
          doc-find-containing-paren
          doc-get-symbols
          get-definition-by-id
-         format!)
+         format!
+         doc-full-tokens)
+
--- a/doc.rkt
+++ b/doc.rkt
@@ -349,9 +349,9 @@
 ;;
 ;; for the first token, its previous token is defined as a zero length fake token which
 ;; has line number 0 and character position 0.
-(define (token-encoding doc token prev-pos)
-  (define-values (line ch) (doc-line/ch doc (SemanticToken-start token)))
-  (define-values (prev-line prev-ch) (doc-line/ch doc prev-pos))
+(define (token-encoding editor token prev-pos)
+  (match-define (list line ch) (send editor pos->line/char (SemanticToken-start token)))
+  (match-define (list prev-line prev-ch) (send editor pos->line/char prev-pos))
   (define delta-line (- line prev-line))
   (define delta-start
     (if (= line prev-line)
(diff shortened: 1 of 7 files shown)


doc.rkt
```diff
@@ -315,6 +316,44 @@
...
+(define (doc-range-tokens doc path pos-start pos-end)
+  (define tokens (collect-semantic-tokens (Doc-text doc) (uri->path path)))
+  (define tokens-in-range
+    (filter-not (λ (tok) (or (<= (SemanticToken-end tok) pos-start)
+                             (>= (SemanticToken-start tok) pos-end)))
+                tokens))
+  (for/fold ([result '()]
+             [prev-pos 0]
+             #:result (let ()
+                        (flatten (reverse result))))
```
dannypsnl, line 350: Is this `(let () X)` cannot be `X`?

doc.rkt
```diff
@@ -315,6 +316,44 @@
...
+  (define indexes (indexes-where *semantic-token-modifiers*
+                                 (λ (m) (memq m (SemanticToken-modifiers token)))))
+  (for/sum ([index indexes])
+    (expt 2 index)))
+
+(define (token-encoding doc token prev-pos)
+  (define-values (line ch) (doc-line/ch doc (SemanticToken-start token)))
+  (define-values (prev-line prev-ch) (doc-line/ch doc prev-pos))
+  (define delta-line (- line prev-line))
+  (define delta-start
```
dannypsnl, line 332: Probably taken spec explanation to here

>    at index 5*i - deltaLine: token line number, relative to the start of the previous token
>    at index 5*i+1 - deltaStart: token start character, relative to the start of the previous token (relative to 0 or the previous token’s start if they are on the same line)

doc.rkt
```diff
@@ -315,6 +316,44 @@
...
                    #:newText new-text)]))
 
+(define (token-type-encoding token)
+  (index-of *semantic-token-types* (SemanticToken-type token)))
+
+(define (token-modifier-encoding token)
+  (define indexes (indexes-where *semantic-token-modifiers*
+                                 (λ (m) (memq m (SemanticToken-modifiers token)))))
+  (for/sum ([index indexes])
+    (expt 2 index)))
```
dannypsnl, line 326: Why is $2^i$? Do you know where is the spec mention this?

struct.rkt
```diff
@@ -129,6 +129,20 @@
...
+(struct SemanticToken
+  (start end type modifiers)
+  #:transparent)
+
+(define *semantic-token-types*
+  '(variable
+     function
+     string
+     number
+     regexp))
```
dannypsnl, line 141: Is this list order irrelevant?

tests/lifecycle/init_resp.json
```diff
@@ -25,6 +25,22 @@
...
+                "legend": {
+                    "tokenModifiers": [
+                        "definition"
+                    ],
+                    "tokenTypes": [
+                        "variable",
+                        "function",
+                        "string",
+                        "number",
+                        "regexp"
```
dannypsnl, line 40: It seems we better generate this JSON via `*semantic-token-types*`

dannypsnl (changes requested):

doc.rkt
6cdh: It's in [semantic token legend section](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokensLegend) :

> Since a token type can have n modifiers, multiple token modifiers can be set by using bit flags, so a tokenModifier value of 3 is first viewed as binary 0b00000011, which means [tokenModifiers[0], tokenModifiers[1]] because bits 0 and 1 are set.

Modifier encoding of a token is a bit flag of its modifiers. Intuitively, the code is

```
(for/fold ([flag 0])
          ([index indexes])
  ; C syntax
  flag | (1 << index)
)
```

`(for/sum ([i indexes]) (expt 2 i))` is a simple equivalent implementation. I should mention this in comment.

doc.rkt
6cdh: It should be `X`. I inserted a `println` to debug in there but forgot to restore it.

doc.rkt
6cdh: yes

struct.rkt
6cdh: The list order is irrelevant.

The client get this list (in this order) from server ability during initialize handshake. Then it uses this list to decode the server messages.

The different order of this list produces different result of encoding. But it still works without change other code.

I should mention this in comment.

tests/lifecycle/init_resp.json
6cdh: yes

6cdh:
This PR has a serious problems, causes latency problem and makes other requests slower to response because it does not run concurrency.

6cdh:
I have no idea how to make it runs concurrently. That method needs to access `doc-text` which is a `lsp-editor%` instance. It's basically a text buffer implementation. But later requests might modify that text buffer. So we need to lock it, then it leaves no room to runs the semantic token request concurrently.

dannypsnl:
How about let others task can interrupt "semantic tokens"?

6cdh:
It needs to access the text buffer during the whole process.

It obtains the text contained in the text buffer, calculate the tokens, then also use the text buffer class method to convert the token positions from character position format into line/character position format.

If it got interrupted in the middle of calculate tokens then the text buffer might got updated and produce unpredictable result.

6cdh:
I found we can use line/character style position directly after I states the whole process before.

And there is another problem that if people types quickly, then it can produce multiple requests. I guess we can always process only one request, and let new request replace the old processing one, like how we did for the traverse that send diagnostics.

6cdh:
It simply copy the text buffer and use it during processing.

It has usable latency now, although the fan is running loudly.

6cdh:
The design that new incoming task replace old running task can't apply. Because client can reuse the result of old requests, for example, insert some whitespaces, then client can calculate the offset and keeps old content highlight. But if the interrupted task returns an empty result, then the highlight would lost. The interrupted task can't just dropped and don't return, because client might wait that request forever and cause resource leak.

I think this PR is completed now.

dannypsnl (approved):

jeapostrophe:
Thank you!

PR #135, opened 2024-09-22
Improve semantic token latency, highlight function and modified variables

6cdh:
This PR added highlight for function and modified variables, and improved expand performance by 30% to 10x.

The details as follows:

* For typed racket, replace its language line with `no-check` language line before expand. It avoid type check and reduce at least one second for each expand. If error happens because of the replace, it would fallback to normal language line.
* Cache namespace. It optimizes for a common scene when people only concentrates on a single file, and don't open other files during the period. In this case, it caches and reuses the namespace of last expand.

This [snippet](https://gist.github.com/6cdh/7d22286f4bd3c5bbaf9375825861cf33) is a demo. For example:

```shell
$ racket fast-expand.rkt path/to/racket-langserver/main.rkt
expand
cpu time: 2196 real time: 2196 gc time: 881
expand2
cpu time: 191 real time: 191 gc time: 71
```

Here the second expand is faster than the first expand by reuses the namespace.

And [a typed racket file](https://github.com/racket/typed-racket/blob/master/typed-racket-test/gui/succeed/racket-esquire.rkt) without the nocheck replacement:

```shell
$ racket fast-expand.rkt path/to/typed-racket/typed-racket-test/gui/succeed/racket-esquire.rkt
expand
cpu time: 2244 real time: 2455 gc time: 792
expand2
cpu time: 1305 real time: 1305 gc time: 392
```

with nocheck replacement:

```shell
$ racket fast-expand.rkt path/to/typed-racket/typed-racket-test/gui/succeed/racket-esquire.rkt
expand
cpu time: 892 real time: 1049 gc time: 324
expand2
cpu time: 226 real time: 226 gc time: 68
```

### DIFF
--- a/expand.rkt
+++ b/expand.rkt
@@ -1,0 +1,97 @@
+#lang racket/base
+
+(provide read-and-expand)
+
+(require racket/match
+         racket/async-channel
+         syntax/modread
+         syntax/parse
+         drracket/check-syntax)
+
+(define *expander-ch* (make-async-channel))
+
+;; (-> input-port path collector% async-channel)
+;; async-channel receives (list (or syntax #f) (or expand-syntax #f))
+;; where `syntax` is the result of read-syntax,
+;; `expand-syntax` is the result of expand.
+;; They can be #f if error happens during processing.
+(define (read-and-expand in path collector)
+  (define ch (make-async-channel))
+  (async-channel-put *expander-ch* (list path in ch collector))
+  ch)
+
+(define (expander)
+  ;; cache namespace for continuous expand request for same uri
+  (define ns (make-base-namespace))
+  (define last-path #f)
+
+  (let loop ()
+    (match-define (list path in out-ch collector) (sync *expander-ch*))
+
+    (define result
+      (cond [(equal? last-path path)
+             (real-expand path in ns collector)]
+            [else
+             (set! ns (make-base-namespace))
+             (real-expand path in ns collector)]))
+
+    (set! last-path path)
+
+    (async-channel-put out-ch result)
+    (loop)))
+
+(define (real-expand path in ns collector)
+  (define-values (src-dir _1 _2) (split-path path))
+  (define-values (add-syntax done)
+    (make-traversal ns src-dir))
+
+  (parameterize ([current-load-relative-directory src-dir]
+                 [current-namespace ns]
+                 [current-annotations collector])
+    (define stx
+      (with-handlers ([(λ _ #t) (λ (exn) exn)])
+        (with-module-reading-parameterization
+          (lambda () (read-syntax path in)))))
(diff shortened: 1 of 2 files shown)


expand.rkt
```diff
@@ -0,0 +1,97 @@
+#lang racket/base
+
+(provide read-and-expand)
```
dannypsnl, line 3: Maybe provide contract?

expand.rkt
```diff
@@ -0,0 +1,97 @@
...
+;; where `syntax` is the result of read-syntax,
+;; `expand-syntax` is the result of expand.
+;; They can be #f if error happens during processing.
+(define (read-and-expand in path collector)
+  (define ch (make-async-channel))
+  (async-channel-put *expander-ch* (list path in ch collector))
+  ch)
+
+(define (expander)
+  ;; cache namespace for continuous expand request for same uri
```
dannypsnl, line 24: Can we cache the namespace via path? i.e. A hashmap that

path1 |-> ns1
path2 |-> ns2
...

6cdh: It's kind of difficult.

The cached namespace would be invalid if the dependency module changes. There also should be a expiration mechanism to remove cache and avoid memory leak.

But it's tricky to know when it would be invalid and when it becomes useless so we can remove it.

dannypsnl: I see, so unless we have a full dependencies tracker system (and must integrate with editor changes), we don't have easy way to find a namespace is invalid. Then the current one is acceptable, thanks!

dannypsnl (approved):

dannypsnl:
I'm wondering how would an invalid namespace behaved? There are two possibility in my mind
1. It throw an exception
2. It still working but produces wrong result

The first one is actually help us to detect it then provide a new one, the second one is harder to handle (wait me modify the snippet & test).

6cdh:
I guess it might throw a exception or silently produce wrong results in different cases.

Because a namespace contains mappings from module names to module bindings. If a module x requires module y, we cache the namespace, then

* change module y's implementation and keep its interface (provided bindings), then it probably produces wrong result.
* change y's interface to provide a subset of old interface, then modify module x. It would produce correct result.
* provide a superset of old interface, then modify module x. It probably throws an undefine error.
* more complex scenario ...

dannypsnl:
Yep, maybe we can create an issue for further discussion, for now the PR is great.

6cdh:
The code to identify function token was complex and has a bug that treat the expression `(if ... proc1 proc2)` in `((if ... proc1 proc2) arg)` as function. I just fixed it.

dannypsnl (approved):

jeapostrophe:
Thanks!

PR #137, opened 2024-10-25
Refactor Services

6cdh:
This is an unexpected big PR.

At the beginning, I found the semantic token highlight service can slow down the whole lsp server, and eats a lot of memory over time. I think that because there are too much semantic token requests to be processed. So I want to move it into `build-trace%` class, make it runs along with check syntax task, and reduce the number of requests.

The problem is that the semantic token request needs to be processed concurrently, and I need to add locks to a lot of code. Most changes in this PR come from the locked code.

To make the concurrent code intuitive and simple, I also considered the design again.

In this design, All requests are classified to two types of nodes:

* text node - denotes the text state of the document. A new document change notification will produce a new text node, and make the old text node invalid.
* query node - denotes the various read only query requests. The query nodes need to access the content of the corresponding text node to do their work.

As the graph shows:

Untitled Diagram

Any new request is either a new text node or a new query node that depends on the current text node.

The text node updates synchronously then runs a check syntax task asynchronously. Most queries also runs synchronously as before. But for async queries (only semantic token service currently), they are scheduled to run before the next text change event or after its check syntax task completed. So there are two events to trigger the process of the async queries.

In other words, a text change will clear all old queries, a successful check syntax task can also clear all old queries.

I also refactored the `build-trace%` class using visitor pattern. And added comment to `methods.rkt` to introduce the design.

The result is that the overall latency is back to previous level and the semantic tokens is faster to response.

The cached expand I added in my previous PR is useless now, but it might be useful to speedup the check syntax task in the future, so I didn't remove it.


lock.rkt
```diff
@@ -0,0 +1,40 @@
+#lang racket/base
+
+;; A Readers Writer Lock
+(struct RWLock
+  (access-sema write-sema readers)
+  #:mutable)
```
dannypsnl, line 6: Maybe the file name should be `rwlock.rkt`, and move into `base/` directory (create one), since the functionality is not directly related to the server.

tests/textDocument/formatting.rkt
```diff
@@ -21,7 +21,8 @@ END
       (let ([notif (make-notification
                     "textDocument/didChange"
                     (hasheq 'textDocument
-                            (hasheq 'uri uri)
+                            (hasheq 'uri uri
+                                    'version 0)
```
dannypsnl, line 25: How do we decide this number? According to spec

> The version number of this document (it will increase after each change, including undo/redo).

Seems it need to count?

dannypsnl (approved):

tests/textDocument/formatting.rkt
6cdh: Yes. But the version number is sent by the client. We don't need to concentrate it. We only need to make sure the version of a new `build-trace%` match its text version.

dannypsnl: I see.

dannypsnl (approved):

jeapostrophe:
This is a pretty big change... are there any more comments from contributors before I merge?

6cdh:
It seems no more comments. I think it's time to merge this.

Issue #138, opened 2025-01-06
`(require malt)` crash langserver


fatboyzz:
```racket
#lang racket
(require malt)
(tensor? #[1 2 3])
```
`malt` is a package from book 'The Little Learner'. 
And the above code works with `Racket: Run file in terminal` and return a `#t`.

But the intellisence seems dead.
Move mouse to identifier `tensor?` shows `Loading ...`.

Delete code from `malt`, add some regular racket code. 
The langserver keeps showing `Loading ...`.

Restart the langserver by toggling extension.
The langserver is back.

6cdh:
The lsp server works and prints the results to stdout, but the editor does not receive it for some unknown reasons.

My debug shows the issue come from this line:

https://github.com/themetaschemer/malt/blob/92a3e63aebf8f0801cffdb6b8a44c199d7973148/impl.rkt#L3

```racket
(require (for-syntax "impl-loader.rkt"))
```

where `for-syntax` is the reason. But I don't know how this happened.

6cdh:
OK, the reason is that `expand` can print something to stdout sometimes. `parameterize` the current output port to a new output string should fix it.

dannypsnl:
https://github.com/themetaschemer/malt/blob/92a3e63aebf8f0801cffdb6b8a44c199d7973148/impl-loader.rkt

```racket
; ...
(when (not (settings))
  (init-settings)
  (println "settings=")
  (pretty-print (settings)))
; ...
```

wow......I think I know why now, then yes overwrite output port should work.

fatboyzz:
It works now, thanks.

PR #139, opened 2025-01-30
fix side effects of `expand` that would print something and messes with stdout

6cdh:
fix https://github.com/jeapostrophe/racket-langserver/issues/138 and similar issues.


jeapostrophe:
Thank you!

shhyou:
How about [open-output-nowhere](https://docs.racket-lang.org/reference/port-lib.html#%28def._%28%28lib._racket%2Fport..rkt%29._open-output-nowhere%29%29)?

6cdh:
yes, it should be `open-output-nowhere`, I didn't know it exists.

PR #144, opened 2025-02-04
supports notification `workspace/didRenameFiles`

### LINKED ISSUE #142: support `workspace/didRenameFiles` notification
The idea is, in this case we just rebind the safe-doc to new key, no trace need to run.

spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename

dannypsnl:
resolve #142

## @6cdh's suggestion

- [x] `Doc` structure contains the URI, therefore, needs to update it
- [x] `scheduler.rkt` (I need to look into it)
- [x] check the new file is a racket file
    - my plan is simply checking `newUri` is `*.rkt`


6cdh:
`Doc` struct contains its uri, so we also need to change that (in `doc.rkt`).

The queries of the old document in queue also need to be handled (maybe just clear them) (in `scheduler.rkt`).

The new file is probably no longer a racket file, for example, someone create a racket file, but after a second, they realize it should be another language, then they rename it. So we need to check if the new file is a racket file.

The most difficult part is, we probably need to consider other files that depends on the old file.

dannypsnl:
@6cdh About dependency relation between files, I thought this is a big refactoring that should be another PR, and maybe a plan, what do you think?

6cdh:
I think so.

dannypsnl:
@6cdh all suggestions are applied.

workspace.rkt
```diff
@@ -11,6 +20,23 @@
...
+
+    ; remove all awaiting internal queries about `old-uri`
+    (clear-old-queries/doc-close old-uri)
+
+    (if (string-suffix? new-uri ".rkt")
+      (let ([safe-doc (hash-ref open-docs (string->symbol old-uri) #f)])
+        ; `safe-doc = #f` should be rarely happened.
+        ; we simply give up to handle it, let's trust LSP client will send others request about analysis this file.
+        (when safe-doc
+          (doc-update-uri! safe-doc new-uri)
```
6cdh, line 36: `safe-doc` is a `SafeDoc` struct contains a `Doc` struct and a rwlock, the `Doc` struct stores its information. So we need to acquire its lock like this:

```racket
(with-write-doc safe-doc
  (lambda (doc)
    (doc-update-uri! doc new-uri)))
```

6cdh (approved):

PR #148, opened 2025-05-21
supports Rhombus files

### LINKED ISSUE #147: Support Rhombus (`*.rhm`) files
Magic Racket (for VS Code) recently gained syntax highlighting support for Rhombus (`*.rhm`) files. The extension also tries to ask the langserver for help with these files, but doesn't seem to get replies.

Is there something server side that needs to be extended to also support `*.rhm` files? Perhaps this capability needs updating:

https://github.com/jeapostrophe/racket-langserver/blob/a37eba78438de72894bf7be42d0da797592d86bf/methods.rkt#L201-L203

Does the langserver already support `*.rhm` in other editors? If yes, then I probably just need to fix something on the Magic Racket side.

I noticed that if you rename a `*.rhm` file to `*.rkt`, then langserver works as expected, so it seems like something is filtering out `*.rhm` files... I'm just not sure yet whether the filtering / bug to be fixed is client or server side.

dannypsnl:
resolve #147

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Next, install the package via `raco`:
 raco pkg install racket-langserver
 ```
 
-Once it is installed, you can configure your editor to use a custom LSP client for Racket files (`.rkt`), and set the command for the custom client to
+Once it is installed, you can configure your editor to use a custom LSP client for Racket/Rhombus files (`.rkt`/`.rhm`), and set the command for the custom client to
 
 ```
 racket -l racket-langserver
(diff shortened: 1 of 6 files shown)


dannypsnl:
I think maybe we could have a list of suffices the server should support, since Racket could have many options?

6cdh (approved):

6cdh:
I agree. We can also make it configurable from outside (config file, command line arguments or something).

jryans:
Yes, would be good to support a configurable list... Not quite sure which ways to configure (file, arguments, etc.) would be best, so that's probably better left to separate issue / PR where we can debate it first.

Thanks for working on this! :smile:

benknoble:
Re: suffixes, see https://github.com/benknoble/vim-racket/issues/14

PR #150, opened 2025-05-22
`compiler/module-suffix` has enough util to do this without any configuration

### LINKED ISSUE #149: Configurable supported suffices list
We currently hard coded what a suffix will be handled by the server, better to have a configurable file to store this list.

dannypsnl:
resolve #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Next, install the package via `raco`:
 raco pkg install racket-langserver
 ```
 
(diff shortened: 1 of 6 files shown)


jeapostrophe (approved):

benknoble:
It would help me, personally, and perhaps others, if commit messages were less vague and more descriptive: a typical formula is

```
short summary

Describe a problem with the current code as it is, explaining why it is a problem or why a change is desirable. 

Give commands to the code to change to the problem.
```

Issue #40, opened 2021-04-03
Resyntax integration


jackfirth:
In the long term, I would like to integrate [Resyntax](github.com/jackfirth/resyntax/) with this language server so that Racket users can get helpful suggestions for improving their code directly in the editor. I don't think Resyntax is stable enough for this yet, but it's a goal I'm working towards. Does this seem like a good idea to folks?

jeapostrophe:
Absolutely. My hope is that the Racket LSP will provide all of the same kinds of things that other LSPs do and I think Resyntax is the thing in the Racket world that fits the mold.

spdegabrielle:
@jackfirth I think resyntax might be _stable enough_ 😁

dannypsnl:
I make a small exploration
```racket
#lang racket/base
(require resyntax
         resyntax/default-recommendations
         resyntax/private/source
         rebellion/base/range
         rebellion/base/comparator
         rebellion/collection/range-set)

(define all-lines (range-set (unbounded-range #:comparator natural<=>)))

(resyntax-analyze
  (file-source "/Users/dannypsnl/tr-notes/tr/private/next.rkt")
  #:suite default-recommendations
  #:lines all-lines)
```

I think this is almost what LSP expecting, we only need to transform its output structures to LSP structures.

cc @6cdh

jackfirth:
I think you'll also have to output some `Diagnostic` messages too.

6cdh:
Yes. We need to add a new function that process integration, send diagnostic and code action to [diagnostic.rkt](https://github.com/jeapostrophe/racket-langserver/blob/master/service/diagnostic.rkt).

I also expanded @dannypsnl code a bit, it should print all information we need to implement the integration:

```racket
#lang racket/base

(require resyntax
         resyntax/default-recommendations
         resyntax/private/source
         resyntax/private/refactoring-result
         resyntax/private/string-replacement
         rebellion/base/range
         rebellion/base/comparator
         rebellion/collection/range-set
         racket/string)

(require racket/pretty)

(define all-lines (range-set (unbounded-range #:comparator natural<=>)))

(define result-set
  (resyntax-analyze
    (file-source "or.rkt")
    #:suite default-recommendations
    #:lines all-lines))

(define results (refactoring-result-set-results result-set))

(define (extract-result result)
  (define sr (refactoring-result-string-replacement result))
  (pretty-display sr)
  (define info
    (list
[...]

PR #153, opened 2025-07-08
Add resyntax integration

6cdh:
This is a initial implementation for #40 .

Some screenshot:

Screenshot 2025-07-08 232519

Screenshot 2025-07-08 232529

Screenshot 2025-07-08 232544

Todo:

- [x] Improve quick fix description
- [x] Implement error handling for unsupported languages
- [x] Incompatibility for old Racket version
- [x] Add a setting to allow editor to enable or disable
- [x] Add test
- [x] Improve latency

For code, I added a new function `walk-text` which provides updated text string to services, it is called every time `check-syntax` succeeds.

### DIFF
--- a/service/dynamic-import.rkt
+++ b/service/dynamic-import.rkt
@@ -12,7 +12,11 @@
                               (void))])
         (dynamic-require mod 'name logging-fail-thunk)))))
 
-(define-syntax-rule (dynamic-imports mod names ... fail-thunk)
+(define-syntax-rule (dynamic-import-mod mod names ... fail-thunk)
   (begin
     (import-once mod names fail-thunk) ...))
+
+(define-syntax-rule (dynamic-imports (mod names ...) ... fail-thunk)
+  (begin
+    (dynamic-import-mod mod names ... fail-thunk) ...))
 
--- a/service/dynamic-import.rkt
+++ b/service/dynamic-import.rkt
@@ -1,0 +1,18 @@
+#lang racket/base
+
+(provide dynamic-imports)
+
+(define-syntax-rule (import-once mod name fail-thunk)
+  (define name
+    (let ([logging-fail-thunk (λ ()
+                                (eprintf "symbol '~a' from module '~a' fail to load.\n" 'name mod)
+                                (fail-thunk))])
+      (with-handlers ([exn? (λ (e)
+                              (logging-fail-thunk)
+                              (void))])
+        (dynamic-require mod 'name logging-fail-thunk)))))
+
+(define-syntax-rule (dynamic-imports mod names ... fail-thunk)
+  (begin
+    (import-once mod names fail-thunk) ...))
+
--- a/service/dynamic-import.rkt
+++ b/service/dynamic-import.rkt
@@ -5,7 +5,7 @@
 (define-syntax-rule (import-once mod name fail-thunk)
   (define name
     (let ([logging-fail-thunk (λ ()
-                                (eprintf "symbol '~a' from module '~a' fail to load.\n" 'name mod)
+                                (log-info "symbol '~a' from module '~a' fail to load." 'name mod)
                                 (fail-thunk))])
       (with-handlers ([exn? (λ (e)
                               (logging-fail-thunk)
--- a/workspace.rkt
+++ b/workspace.rkt
@@ -11,23 +11,23 @@
 (require "open-docs.rkt")
 
 (define-json-expander FileRename
-  [oldUri string?]
-  [newUri string?])
-(define-json-expander  RenameFilesParams
-  [files (listof hash?)])
+                      [oldUri string?]
+                      [newUri string?])
+(define-json-expander RenameFilesParams
+                      [files (listof hash?)])
 
 (define-json-expander WorkspaceFolder
-  [uri string?]
-  [name string?])
+                      [uri string?]
+                      [name string?])
 (define-json-expander WorkspaceFoldersChangeEvent
-  [added (listof hash?)]
-  [removed (listof hash?)])
+                      [added (listof hash?)]
+                      [removed (listof hash?)])
 
 (define-json-expander FileEvent
-  [uri string?]
-  [type exact-positive-integer?])
+                      [uri string?]
+                      [type exact-positive-integer?])
 (define-json-expander DidChangeWatchedFilesParams
-  [changes (listof hash?)])
+                      [changes (listof hash?)])
 
 (define workspace-folders (mutable-set))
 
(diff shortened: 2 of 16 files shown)


6cdh:
And what do you think user defined rules? Should we support it? And how to support it?

dannypsnl:
I have a similar question, if user didn’t like a certain suggestions, where can they disable it?

It seems we also need magic-racket they give us those configurations?

dannypsnl:
misc

- [ ] test rhombus 
- [ ] update CI

jackfirth:
For user-defined rules, I don't think supporting that should be the editor's responsibility. That should be integrated into Resyntax proper - possibly through some mechanism where some `info.rkt` config tells Resyntax where to find the rules.

jackfirth:
@dannypsnl Resyntax's goal is to not provide suggestions that people frequently want to disable. So far, that's mostly been working. If there are suggestions that people don't find to be helpful, I would rather they file issues in the Resyntax github repository than disable them themselves. But see also some of my thoughts for suppressing suggestions within Resyntax itself: jackfirth/resyntax#436.

jryans:
First off, this a neat feature, thanks for working on it! :smile: 

I do think some means to at least to disable suggestions entirely is needed, as not everyone will want them to appear.

The main pathways I can imagine are:

- editor setting which is passed along to the lang server
- some resyntax-layer config in a file somewhere

I can also imagine some people wanting more fine-grained control beyond just off / on, but that's perhaps a future refinement.

Looking at other language ecosystems, Rust Analyzer [takes the first path](https://rust-analyzer.github.io/book/configuration.html) of having quite a few editor-level settings which are then passed to the lang server. IIRC, we don't have any concept of that for Racket so far, but it seems like a reasonable way for users to communicate their preferences for tooling features like this.

6cdh:
Racket-langserver now requires minimum version of Racket 8.16, from the previous 7.6. Because resyntax uses treelist and `sequence->treelist`.

6cdh:
I added settings to enable editor control features. By default it's this:

```json
{
    "resyntax": {
        "enable": true
    }
}
```

@jryans Please test if it works if you have time.

dannypsnl:
> For unsupported language files, resyntax can skip them automatically, so we don't need to take any action.

If we create another function, e.g. `resyntax-analyze-lsp` which use different `skip` function?

jryans:
> I added settings to enable editor control features. By default it's this:
> 
> ```json
> {
>     "resyntax": {
>         "enable": true
>     }
> }
> ```
> 
> @jryans Please test if it works if you have time.

Thanks, I will endeavour to take a look, but it may take me a while (recent illness plus too many things to do), so please don’t feel you need to wait on me.

jackfirth:
Resyntax has so far been built on the assumption it's running in a CI environment where it's easy to use the most recent Racket version. You won't be able to get Resyntax working on an older Racket version from the outside. It would need some fairly extensive work done on Resyntax itself.

6cdh:
@dannypsnl

> You can use [dynamic-require](https://docs.racket-lang.org/reference/Module_Names_and_Loading.html#%28def._%28%28quote._~23~25kernel%29._dynamic-require%29%29) to detect if required stuff is there or not

Resyntax probably also uses other new features, and only detect treelist is not enough. But maybe we can use `dynamic-require` to load Resyntax dynamically. If Resyntax exists and load correctly, enable it, otherwise disable it. I don't know if it's possible, probably need some experiment.

> If we create another function, e.g. resyntax-analyze-lsp which use different skip function?

I don't understand your meaning. In this code

```racket
(define result-set
  (resyntax-analyze
    text-source
    #:suite default-recommendations
    #:lines all-lines))
```

If the `text-source` is unsupported language, `resyntax-analyze` skips the source and returns a empty result set. So we don't need to handle the unsupported case explicitly.

dannypsnl:
@6cdh Indeed, dynamic require resyntax is enough.

I mean the stderr output you mentioned, that’s from an resyntax internal logging function.

> But every time resyntax skips a file, it always print this log to stderr

workspace.rkt
```diff
@@ -77,3 +81,11 @@
   (when (regexp-match (get-module-suffix-regexp) uri)
     (clear-old-queries/doc-close uri)
     (hash-remove! open-docs (string->symbol uri))))
+
+(define (didChangeConfiguration params)
+  (match-define (hash-table ['settings settings]) params)
+
+  (define key '(resyntax enable))
+  (when (jsexpr-has-key? settings key)
+    (set-resyntax-enabled! (jsexpr-ref settings key #t))))
```
dannypsnl, line 90: I think `jsexpr-ref` only takes two arguments? They are `jsexpr` and `keys` (access chain), and no fallback value.

service/dynamic-import.rkt
```diff
@@ -0,0 +1,18 @@
...
+                                (eprintf "symbol '~a' from module '~a' fail to load.\n" 'name mod)
+                                (fail-thunk))])
+      (with-handlers ([exn? (λ (e)
+                              (logging-fail-thunk)
+                              (void))])
+        (dynamic-require mod 'name logging-fail-thunk)))))
+
+(define-syntax-rule (dynamic-imports mod names ... fail-thunk)
+  (begin
+    (import-once mod names fail-thunk) ...))
```
dannypsnl, line 17: I think these not yet need to be out of `resyntax.rkt`, until we really need to reuse them.

dannypsnl (changes requested):

workspace.rkt
6cdh: Yes, I forgot to remove that `#t`.

jackfirth:
> > If we create another function, e.g. resyntax-analyze-lsp which use different `skip` function?
> 
> Oh you mean this [skip function](https://github.com/jackfirth/resyntax/blob/63d101fda8b58676022fa857f9b7ecc178efdb14/main.rkt#L154)
> 
> Yes, I agree.

That's using a normal racket [logger](https://docs.racket-lang.org/reference/logging.html#%28tech._logger%29) with the topic `'resyntax`, so if you want to suppress that output you can do so through racket's usual logger config mechanisms, like the [`PLTSTDERR`](https://docs.racket-lang.org/reference/logging.html#%28idx._%28gentag._451._%28lib._scribblings%2Freference%2Freference..scrbl%29%29%29) environment variable.

service/dynamic-import.rkt
6cdh: This file is small, but a extra file is not harmful. I think it's better to split it up.

6cdh: Also we will need it in a test, that test is only enabled when Resyntax exists.

dannypsnl: I see.

dannypsnl (approved):

jackfirth:
> I prefer it does not show logs by default and only shows by given this environment variable or some explicit signals.

You can change the value of `PLTSTDERR` to do that. It defaults to `error`, but if the racket langserver sets it to `error fatal@resyntax` before starting the racket process, that should silence it.

6cdh:
> the problem is that it logs these errors when analyzing files in unsupported languages, correct?

Yes, that's the problem.

6cdh:
There are two tasks at the end.

The first one is that resyntax adds a delay to the `check-syntax` function https://github.com/jeapostrophe/racket-langserver/blob/29089574b5f07c48caaa3bc833a8782cb78bb3d0/check-syntax.rkt#L139 Given `methods.rkt` in this project, this is the result:

```
(expand original-stx)
cpu time: 1993 real time: 1993 gc time: 906
(add-syntax stx)
cpu time: 155 real time: 155 gc time: 23
(send new-trace walk-stx original-stx stx)
cpu time: 31 real time: 31 gc time: 1
(send new-trace walk-text text)
cpu time: 4959 real time: 4959 gc time: 79
```

The first three expressions are the main overhead in `check-syntax`, the last expression is the new added resyntax pass. So it adds 2x to 5x time, which is too long especially the `expand` is already very slow.

Make it asynchronous and cancellable, like how the `check-syntax` function is executed, should solve it. It would take some time.

Another task is that we are using [DidChangeConfiguration Notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration) to get updated settings from client when users change settings or when lsp startup. But they have a new [Configuration Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration) API that let server send request to client to get settings.

Personally I like the old API, but some editors like neovim only supports the new API. Supports the new interface would take some time because racket-langserver cannot send request to client right now.

dannypsnl:
We can have a new issue for server request to client this new ability.

6cdh:
The async version looks weird but works. This PR looks fine and complete.

> We can have a new issue for server request to client this new ability.

Yes. Maybe another PR to process [Configuration Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration) API. This is not very relevant to resyntax integration.

dannypsnl (approved):

jeapostrophe:
Thank you everyone.

PR #156, opened 2025-08-28
Support configuration request

dannypsnl:
1. support server request via JSON RPC

TODO
- [ ] use this method to request configuration

### DIFF
--- a/methods.rkt
+++ b/methods.rkt
@@ -49,6 +49,7 @@
 ;; Mutable variables
 (define already-initialized? #f)
 (define already-shutdown? #f)
+(define response-handlers (make-hash)) ; Each request sent by server will record its response handler here
 
 ;;
 ;; Dispatch
@@ -76,6 +77,11 @@
     [(hash-table ['method (? string? method)])
      (define params (hash-ref msg 'params hasheq))
      (process-notification method params)]
+    [(hash-table ['jsonrpc "2.0"]
+                 ['id id]
+                 ['result result])
+     (define handler (hash-ref response-handlers id))
+     (handler result)]
     ;; Batch Request
     [(? (non-empty-listof (and/c hash? jsexpr?)))
      (for-each process-message msg)]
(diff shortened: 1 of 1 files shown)


dannypsnl:
https://prefab.cloud/blog/lsp-language-server-configuration-example/

6cdh:
About the test, I think just add a new function `(client-send-request-and-handle lsp req handler)` in [client.rkt](https://github.com/jeapostrophe/racket-langserver/blob/master/tests/client.rkt) should work. And wait for all handlers complete before [shutdown](https://github.com/jeapostrophe/racket-langserver/blob/f8eb192597f36c1f2801995822cd0a492c60f607/tests/client.rkt#L47) or error if some handlers are not called within a timeout. Or just simply `(client-send-request-and-wait lsp req)`, and return the corresponding response.

6cdh:
Most settings should be the same across files. The file specific settings should be small. And our settings are also very small right now.

And `workspace/configuration` cannot detect client settings change. `workspace/didChangeConfiguration` can do. We can just request settings once and cache. Let `workspace/didChangeConfiguration` process runtime settings change.

The specification seems expects server to cache `workspace/configuration`:

> If the server still needs to react to configuration changes (since the server caches the result of workspace/configuration requests) the server should register for an empty configuration change using the following registration pattern ...

jryans:
Thanks for working on this...! :smile: I haven't dug into the details of how this normally works in VS Code (and other editors). Is there anything Magic Racket can do to help by e.g. communicating configuration info in a specific way...?

dannypsnl:
NOTE: https://github.com/rust-lang/rust-analyzer/issues/4267

6cdh:
So the configuration model looks like this?

* When lsp server starts up or a document is opened in editor, the server sends a `workspace/configuration` request to editor and caches received settings.
* When users change settings in editor, the editor sends a `workspace/didChangeConfiguration` notification to server. It's either the new settings (old style) or empty (new style). In old style, server updates these settings. In new style, this notification invalids all cached settings, and server then sends a `workspace/configuration` to fetch new settings.

text-document.rkt
```diff
@@ -73,12 +75,24 @@
...
 
+(define-json-expander ConfigurationItem
+  [scopeUri string?]
+  [section string?])
+
+(define (fetch-configuration uri)
+  (send-request 0 "workspace/configuration"
+    (hasheq 'items (list (ConfigurationItem #:scopeUri uri #:section "racket-langserver")))
+    (λ (configuration)
+      ; TODO: apply configuration
```
6cdh, line 86: We can split the logic from function `didChangeConfiguration` in `workspace.rkt`, and reuse it here.

dannypsnl:
It seems the new model is what you said, but I also found the client part can customize these (but I haven't see any client do this, welcome any new information about this), so what's the default behavior seems depends on certain editor.

6cdh (approved):

PR #162, opened 2025-11-23
Use parallel thread

dannypsnl:
I will use this branch for a few days before I make PR ready.


dannypsnl:
Emmm, obviously not as easy as I thought.

dannypsnl:
@shhyou Indeed, that's a problem.

I will close this PR, because the model, expansion is only going for the current editing file, where parallel thread can't help.

Also, I shall not force an upgrade, I think we can
1. wait until Racket users all use >=9.0
2. or create an experimental branch, creates a complete different interaction to documents

PR #165, opened 2025-12-20
Don't clear completions when reset document

### LINKED ISSUE #163: textDocument/completion behaving strangely
I've got a few possible issues here, and I'm not sure what's at the root, but this is the information I've been able to collect.

I noticed all this because my LSP client receives no completions from the langserver. I used to get accurate completions.

## Symptoms

Consider the file https://github.com/benknoble/frosthaven-manager/blob/main/defns/level.rkt (for example, `raco pkg install --clone fhm https://github.com/benknoble/frosthaven-manager`).

If I navigate to the end of the file and start appending new text, for example `le`, then trigger completion, then I find the following[^2]:

- completion params are `#hasheq((position . #hasheq((character . 4) (line . 45))) (textDocument . #hasheq((uri . "file:///Users/Knoble/code/frosthaven-manager/defns/level.rkt"))))`
- pos for guessing tokens is 1261
- token is `le`, but no completions are found (shouldn't we find `level-info` and so on?)

Same thing happens for `ma`, which should complete at least `max`, `match`, etc?

[^2]: PS Occasionally I get an "index out of range error" which makes it seem like the open document doesn't have the contents of the Vim buffer and is operating from disk, but that might be a client  issue, and I haven't been able to reproduce it with reliability. It might be related to the "sometimes the document text isn't right" issue.

Now, I go to the middle of the file… let's take the `list-ref`, remove the `f`, and trigger completion after the `e`, then we get:

- completion params are `'#hasheq((position . #hasheq((character . 10) (line . 40))) (textDocument . #hasheq((uri . "file:///Users/Knoble/code/frosthaven-manager/defns/level.rkt"))))`
- pos for guessing tokens is 1172
- _sometimes_ the document text doesn't have the `list-re`[^1], and so the "token" becomes `level-`, and no completions are returned (I think we should still get some?). It seems to me this happens most often when I restart the server altogether and trigger completion, so it might be some kind of client issue, but I'm not sure. I don't know how this code cooperates (or not) with clients to look at buffer contents, and it's possible Vim+ALE does something strange with the region of text between where I triggered completion and where ALE determines the completion should start matching.
- It seems if I wait long enough, then trigger completion, the document text is correct, but the token is `(list-re`, and no completions are returned. Even removing the opening paren in such a situation (so the token is `list-re`) gives no results, and same if the token is `list`.

Now ironically, it seems that if I put the `f` back in and trigger completion, then the token is `(list-ref`, and the langserver responds with [a large list of possible completions](https://gist.github.com/benknoble/51402e434d218c7e3ef421abc671a5ae) (which my client of course filters down to "only possible completion is `list-ref`).

That also means that if I trigger completion with an _empty_ token (say, on any blank line), I get results! But I could swear that even last time I was regularly programming in Racket (last fall?) completion worked anywhere. However, if I start a new line (`o` in Vim) then trigger completion, the same empty token yields no completions?

Also, in the results, I get some of the expected definitions from that module, but I don't get `max` or `match` (I do get `prop:legacy-match-expander` of all things, in addition to the obvious `max-level` and `max-players`). So something is going very wrong here.

## Device information

- Old macbook, but hardward shouldn't matter (?)
- Client: dense-analysis/ale@ca1da76d5e91e42f676654905c0c7b6d074b8068 (for Vim)
- Racket: v8.17 CS
- langserver: a9297eddaa3f4b7689e4d1594bf5a3e44cfaaa9a (+ debugging modifications)

## Debugging strategy

I thought it might be my client, so I'd added some tracing prints there. I can continue to do so, but I now suspect that at least something is funky in the server. If the server can be exonerated, I can continue to trace the client and report bugs there, but I don't _think_ my other servers are showing the same problem. Rust, for example, works just fine.

I've also added some tracing prints to the Racket server (compile with `raco setup -l racket-langserver` afterwards so that things aren't bogged down ;) Here's a diff of what I've added so far (adding the `with-handlers` + `debug` print to `(send doc-trace get-completions)` brought things to a halt even with recompilation, but since the final results are empty it must be empty, too).

<details><summary>patch for debugging code (generated with diff, _not_ git diff)</summary>

```patch
diff ./doc.rkt /Users/Knoble/Library/Racket/8.17/pkgs/racket-langserver/doc.rkt
400a401,405
> (define-syntax-rule (d! x ...)
>     (with-output-to-file
>      #:exists 'append "/tmp/langserver"
>      (λ () x ...)))
> 
401a407,408
>   (d! (println `(pos ,pos)))
>   (d! (println `(text ,(send (Doc-text doc) get-text))))
diff ./text-document.rkt /Users/Knoble/Library/Racket/8.17/pkgs/racket-langserver/text-document.rkt
251a252,256
>   (define-syntax-rule (d! x ...)
>     (with-output-to-file
>      #:exists 'append "/tmp/langserver"
>      (λ () x ...)))
>   (d! (println params))
262c267,281
<                    (send doc-trace get-online-completions (doc-guess-token doc pos))))
---
>                    (let ([x
>                           (with-handlers ([exn:fail?
>                                           (λ (e)
>                                             (d!
>                                              (parameterize ([current-error-port (current-output-port)])
>                                                ((error-display-handler)
>                                                 (exn-message e) e)))
>                                             (raise e))])
>                             (send doc-trace get-online-completions
>                                   (let ([x (doc-guess-token doc pos)])
>                                     (d! (println `(guessed-token ,x)))
>                                     x)))])
>                      (d! (println `(online ,x)))
>                      x)))
>          (d! (println `(comp ,completions)))
```

</details>

With this code applied, I can `tail -f /tmp/langserver` in one terminal and trigger completion in another. If there is a better strategy for debugging, please let me know. And please let me know what other information would be helpful to collect.

[^1]: It looks like `(define (get-level-info level)\n  ( level-table level))\n\n` in the string, which… that's not right. I didn't even modify the Vim buffer to look like that.

6cdh:
fix #163

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
       fail-fast: false
       matrix:
         racket-version: ["stable"]
-        racket-variant: ["BC", "CS"]
+        racket-variant: ["CS"]
     steps:
       - uses: actions/checkout@v3.1.0
       - uses: Bogdanp/setup-racket@v1.14


6cdh:
I did a rebase. Please review this again. We can also merge this now.

PR #200, opened 2026-03-21
Add external formatter support

### LINKED ISSUE #178: Support external formatter
Use external formatters to handle formatting request of the protocol. See https://github.com/jeapostrophe/racket-langserver/issues/175#issuecomment-3828383215

6cdh:
This PR switches the existing handwritten formatting logic to external formatters.

Currently, only racket-fixw is integrated because it's error-tolerant, and supports interactive mode.

- For on type formatting, use interactive mode. It indents for empty lines as if there is an atom on these lines based on the context. So if user types `enter`, they are at the indented position, not the beginning of the line.
- For full document and range formatting, use normal mode, that does not indent for empty lines because they are treated as trailing spaces.

Resolve #178

And after this change, we can use a handwritten text buffer to replace `racket:text%` class, and remove the GUI `framework` dependency.

### DIFF
--- a/doclib/doc.rkt
+++ b/doclib/doc.rkt
@@ -9,6 +9,7 @@
          "editor.rkt"
          "../common/path-util.rkt"
          "doc-trace.rkt"
+         "formatting.rkt"
          "internal-types.rkt"
          racket/match
          racket/contract
@@ -16,7 +17,6 @@
          racket/set
          racket/list
          racket/string
-         racket/bool
          racket/dict
          data/interval-map
          syntax-color/module-lexer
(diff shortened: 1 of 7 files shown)


doclib/doc.rkt
```diff
@@ -349,117 +349,32 @@
...
+(define (doc-src-dir doc)
+  (with-handlers ([exn:fail? (λ (_exn) #f)])
+    (define doc-path (uri->path (Doc-uri doc)))
+    (define-values (base _name _must-be-dir?) (split-path doc-path))
+    base))
+
 ;; Shared path for all formatting requests
 (define/contract (doc-format-edits doc fmt-range
-                                   #:formatting-options opts
+                                   #:formatting-options _opts
```
dannypsnl: Since we are not using this parameter, shall we just remove it?

dannypsnl: I see

dannypsnl (approved):

PR #205, opened 2026-03-26
Fix performance test in CI

6cdh:
- Increase time budget in tests
- Use CPU count instead of hardcoded 16 in `raco test`


.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Install Resyntax
         run: raco pkg install --no-docs --auto resyntax
       - name: Running resyntax tests
-        run: raco test -j 16 tests/textDocument/resyntax/
+        run: raco test tests/textDocument/resyntax/
```
dannypsnl, line 59: we don't need parallel here?

dannypsnl:
Can we also update `info.rkt` here? `--check-pkg-deps` should help us to find dependency that no declarations.

.github/workflows/ci.yml
6cdh: Yes, only one file there. And Resyntax integration is stable I think.

6cdh:
I removed the legacy text buffer, because `raco setup` compiles all Racket files in the project. It makes GUI library used when compiling legacy text buffer. I also removed the benchmark script because there is no comparison target now.

tests/lib/editor-test.rkt
```diff
@@ -263,8 +263,8 @@
 
   (define practical-line-count 2000)
   (define practical-line-width 40)
-  ;; 0.001 time budget is enough for recent laptops, 0.005 for CI environments in parallel tests
-  (define practical-time-budget 0.005)
+  ;; 0.001 time budget is enough for recent laptops, 0.01 for CI environments in parallel tests
```
dannypsnl, line 266: What is the first 0.001 for?

6cdh: 0.001 (1ms) is the original time spec, 0.01 for weak CI machines.

dannypsnl: Oh, so that constant is defined elsewhere?

dannypsnl (approved):

tests/lib/editor-test.rkt
6cdh: I'm kind regret it. It should detect CI environment and use different values than local machine. But right now this code is fine and works.

shhyou:
What's the rationale to remove `racket-doc` from the dependency? It's causing `raco setup` problems because building the documentation depends on it.

shhyou:
#189 I think. Running `raco setup --check-pkg-deps --pkgs racket-langserver` reports the missing dependency.

shhyou:
The CI script intentionally skips building the documentation via `--no-docs` so it won't report this dependency.
https://github.com/jeapostrophe/racket-langserver/blob/bafff8b5d5699034567301914a9d40009f4ff72f/.github/workflows/ci.yml#L33-L34

dannypsnl:
We can create an independent yaml to check dependencies.

6cdh:
I forgot that PR totally, and didn't notice what `--no-docs` do. It was my mistake, and should be fixed quickly.

shhyou:
Ah, thanks for the immediate fix!

PR #210, opened 2026-04-25
Add Language detection and dispatch logic

6cdh:
The current LSP server's behavior for non-Racket or non-Lisp languages is unclear. For example, we don't have clear idea what features support what languages. Almost all features target to Racket, but services that operate on expanded code seems more general purpose.

This PR adds language detection and dispatch logic depends on language. And add docs for behavior for various languages.

TODO:

- [x] Add a simple token tree parser (no query interface yet)
- [x] Detect language header with error-tolerance: `#lang ...` (including `#lang reader ...`), `#reader ...`, `(module id mod-path ...)`, if no languag header found, fallback to guess language from special file suffix, like `.rhm`, `.scrbl`
- [x] Maintain a predefined list of languages and their properties. If a recognized language is found in this list, we can apply special logic to them. If a language name is not in this list, it is not treated as a Sexp language, and not a Lisp
- [x] Stop pass indenter to services, as it's not used
- [x] Improve diagnostic messages for missing and malformed language header
- [x] Don't formatting non-Sexp languages
- [x] Let lexer snapshot stores a token tree, add query interface to token tree, and wire structural queries to use it, instead of traversing and matching parenthesis on flat tokens
- [x] Allow semantic highlighting to highlight sexp comment `#;`
- [x] Add language support matrix in README
- [x] Add features.md for detailed docs of features
- [x] Add more languages to predefined list
- [x] Process rktd policy in doc-lang

The language header detection is based on the Sexp token tree parser. I didn't use regex because comments can appear before language line, especially Sexp comment. So it builds a small Sexp token tree before the language header. Then recognize the language, if it's in the predefined list, and is a Sexp language, the token tree parser will continue to parse the rest code (continue parsing logic not yet implemented). For non-Sexp language, its Sexp token tree only contains the code before and including the language header. If the language header is missing, the Sexp token tree parser can potentially read the whole document.

The token tree parser can also support non Sexp, but not yet. Because each non Sexp language can have its special syntax, and make the implementation more complex.


common/path-util.rkt
```diff
@@ -5,15 +5,13 @@
...
-         racket/string
          racket/list
          racket/path)
 
 (define path->uri (compose url->string path->url))
 
 (define (uri->path uri)
-  (cond [(string-prefix? uri "file:") (path->string (url->path (string->url uri)))]
-        [else (uri->path (regexp-replace #rx".*?:" uri "file:"))]))
+  (path->string (url->path (string->url uri))))
```
dannypsnl: just `compose`?

scribblings/racket-langserver.scrbl
```diff
@@ -699,8 +699,8 @@ Exceptions are noted in individual entries.
                       [#:on-type? on-type? boolean? #f])
          (or/c (listof TextEdit?) #f)]{
   Computes formatting edits for the lines covered by @tt{fmt-range}.
-  Returns a list of @racket[TextEdit] values to apply, or @racket[#f] if no
-  indenter is available (e.g., the document lacks a @tt{#lang} line).
+  Returns a list of @racket[TextEdit] values to apply. For documents without
+  a recognized s-expression language, returns an empty list.
```
dannypsnl, line 702: Do we only handle s-expression language here? There are actually many non s-expression languages in Racket system.

6cdh: This is a problem. We shouldn't route non-Sexp languages to fixw or other Sexp formatters, while almost all formatters in Racket community are Sexp formatters. So we have to have a guard in format logic.

I'm thinking three methods:

1. Specialize for common non-Sexp languages when we have dedicated formatters for them, for example, scribble or rhombus.
2. Use the language provided indenter if it provides [`'drracket:indentation`](https://docs.racket-lang.org/tools/lang-languages-customization.html#(part._.Indentation)) procedure. But this indent procedure requires a [`color-textoid<%>`](https://docs.racket-lang.org/syntax-color/color-textoid.html#%28def._%28%28lib._syntax-color%2Fcolor-textoid..rkt%29._color-textoid~3c~25~3e%29%29) argument. Current text buffer does not implement `color-textoid<%>` interface. And it requires some effort to implement it. My attempt shows it needs another 300 lines of complex code. This is still a good option, if we want best support for various languages. Our old handwritten formatter used this method, but used the text buffer that defined in `framework` that is a GUI library.
3. Use method 2, but in a external formatter, so this external formatter becomes a universal indenter for Racket ecosystem. And we just route non S-exp languages to it. We don't route all languages to it because a Sexp formatter usually does more than just indent.

6cdh: I'd like to delay the decision, and avoid to make a `color-textoid<%>` interface implementation here. AI can do it quickly, but review would be pain.

6cdh:
I think this PR is ready to merge.

doclib/service/diagnostic.rkt
```diff
@@ -142,42 +194,45 @@
...
      ;; however it simply doesn't highlight any code.
-     (define silly-range
-       (Range #:start (Pos #:line 0 #:char 0) #:end (Pos #:line 0 #:char 0)))
-     (list (Diagnostic #:range silly-range
+     (list (Diagnostic #:range (first-line-range doc-text)
                        #:severity DiagnosticSeverity-Error
                        #:source "Racket"
                        #:message msg))]
+    ;; If a new kind of expansion failure reaches this point, fail loudly so we
+    ;; can decide how to map it into an LSP diagnostic.
```
dannypsnl, line 235: but the behaviour at this point is still the old treatment?

6cdh: The logic here is expected, I think. What is the problem?

lsp/text-document.rkt
```diff
@@ -262,8 +262,16 @@
     (Range (doc-abs-pos->pos doc current-line-start-pos)
            (doc-abs-pos->pos doc current-line-end-pos)))
 
+  ;; TODO: Gate this sexp-structure lookup to sexp languages, or keep
+  ;; non-sexp documents on current-line formatting only.
```
dannypsnl, line 266: Is this resolved?

doclib/service/diagnostic.rkt
dannypsnl: ok, I see

lsp/text-document.rkt
6cdh: No. In lsp layer, all three formatting methods call a range formatting API at doclib layer. That API is guarded by a sexp language check. To avoid the extra containing form lookup at lsp layer, we will need to add another on-type-formatting API in doclib layer. I'm doing it now.

dannypsnl (approved):

Issue #207, opened 2026-04-02
syntax highlight rktd files


jestarray:
I'm told racket data have the extension `.rktd`, for just data files, as is `.json` to `.js`, however `racket-langserver` nags you if the 1st line isn't a `#lang` definition.

```
Missing or invalid #lang lineRacket
```

We might also want to PR magic racket to support syntax highlighting for `.rktd` files

6cdh:
The read-language is here:

https://github.com/jeapostrophe/racket-langserver/blob/bafff8b5d5699034567301914a9d40009f4ff72f/doclib/check-syntax.rkt#L13-L28

The diagnostic message is produced here:

https://github.com/jeapostrophe/racket-langserver/blob/bafff8b5d5699034567301914a9d40009f4ff72f/doclib/service/diagnostic.rkt#L45-L51

Add a guard to prevent produce this message for `rktd` files should fix this.

dannypsnl:
@jestarray You might like to check #213

dannypsnl:
Complete at #210

PR #213, opened 2026-06-05
stop require `#lang` line for data files

dannypsnl:
related to #207

### DIFF
--- a/doclib/check-syntax.rkt
+++ b/doclib/check-syntax.rkt
@@ -4,6 +4,7 @@
          racket/contract
          racket/logging
          syntax/modread
+         racket/path
          racket/port
          "editor.rkt"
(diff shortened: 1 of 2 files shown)


6cdh:
This is a language policy issue. Can we address it in #210 ? That PR already has a language policy module there, and `.rktd` file handling is in the todo list as well. And some other changes in that PR, for example, not pass indenter to services, which will overwrite some changes here.

That PR was moving slowly, recently I have more time and working on it, so it should complete soon.

I will approve the review, you can merge it if you'd like to resolve it quickly.

6cdh (approved):

dannypsnl:
I think we can have this merged as a quick fix for now, and #210 as a final solution.

PR #215, opened 2026-06-14
sticky scroll for `#lang racket`

### LINKED ISSUE #180: Feature Request: Function sticky scroll header
When you scroll inside a long function (or variable declaration etc.), VS Code keeps the current scope header “stuck” to the top of the editor. So if you're deep of a long function, you'll always know where you are in the code base by looking at the first line

https://youtu.be/oDztOfSbKVg?t=85

dannypsnl:
resolve #180

### DIFF
--- a/common/interfaces.rkt
+++ b/common/interfaces.rkt
@@ -28,6 +28,7 @@
          (json-type-out DocumentHighlight)
          (json-type-out SymbolKind)
          (json-type-out SymbolInformation)
+         (json-type-out DocumentSymbol)
          (json-type-out Hover)
          (json-type-out SignatureInformation)
          (json-type-out SignatureHelp)
@@ -156,6 +157,17 @@
   [name string?]
   [kind SymbolKind]
   [location Location])
+
+;; Hierarchical document symbol. `range` covers the whole form (including its
+;; body) while `selectionRange` only covers the name. `children` is a list of
+;; DocumentSymbol, typed as list? because define-json-struct cannot express
+;; self-referential field types; encoding still recurses via ->jsexpr.
+(define-json-struct DocumentSymbol
+  [name string?]
+  [kind SymbolKind]
+  [range Range]
+  [selectionRange Range]
+  [children list?])
 
 (define-json-struct Hover
   [contents string?]
--- a/doclib/doc.rkt
+++ b/doclib/doc.rkt
@@ -15,6 +15,16 @@
          "lexer.rkt"
          (only-in "lexer/state.rkt"
                   lexer-state-body-forest)
+         (only-in "lexer/token-tree.rkt"
+                  Token-Leaf? Token-Leaf-span
+                  Token-List? Token-List-children
+                  Token-Prefix-Tree?
+                  Token-Forest-nodes
+                  token-node-children
+                  token-node-start
+                  token-node-end
+                  non-skippable-node?
+                  token-leaf-type?)
          "doc-lang.rkt"
          racket/match
          racket/contract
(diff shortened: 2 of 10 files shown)


common/interfaces.rkt
```diff
@@ -157,6 +158,17 @@
...
+;; body) while `selectionRange` only covers the name. `children` is a list of
+;; DocumentSymbol, typed as list? because define-json-struct cannot express
+;; self-referential field types; encoding still recurses via ->jsexpr.
+(define-json-struct DocumentSymbol
+  [name string?]
+  [kind SymbolKind]
+  [range Range]
+  [selectionRange Range]
+  [children list?])
+
```
6cdh, line 171: This is embarrassing, I didn't consider recursive types when designing them.

This is an interesting problem.

I just considered it, the bare identifier should be treated as json types (defined using these DSL macros), not to be Racket predicates or contracts. And I can add another type spec syntax to this DSL, for example, `(pred ctc)`, to express `ctc` is a predicate or contract, that escapes DSL syntax to normal Racket contract. Then do not check if bare identifier is a bounded json types at expand time. This would also produce other smaller problems, this is not in scope of this PR. I will do it in another PR.

doclib/lexer/snapshot.rkt
```diff
@@ -83,6 +83,15 @@
...
 
+(define/contract (lexer-entry-paren-kind entry)
+  (-> LexerEntry? (or/c 'open 'close #f))
+  (case (LexerEntry-type entry)
+    [(open-paren)
+     (and (memv (string-ref (LexerEntry-text entry) 0) '(#\( #\[ #\{)) 'open)]
+    [(close-paren)
+     (and (memv (string-ref (LexerEntry-text entry) 0) '(#\) #\] #\})) 'close)]
+    [else #f]))
+
```
6cdh: BRO you must use LLM for this, they like to do this 😀. This function is literally a wrapper of `LexerEntry-type`, which returns symbols of tokens, means their types. And the more important is, this function has only one usage, in `doc.rkt` line 953, is: `(eq? (lexer-entry-paren-kind ...) 'open)`. This also happens to me, LLMs like to write a new wrapper for lexer token types...

doclib/doc.rkt
```diff
@@ -795,6 +795,236 @@
       #:location (Location #:uri uri
                            #:range range))))
 
+;; Hierarchical Document Symbols: returns a tree of DocumentSymbol.
```
6cdh, line 808: This implementation is over complex. The lexer modules provides two view of tokens. One is a flat stream of tokens, it's basically a vector of tokens. Another one is trees of tokens, it's a token forest.

This implementation uses flat stream, and build a tree, and maintain it. And check if a head element is in a predefined list, if it is, build a `DocumentSymbol` using it and its position. This is over complex, because, you can directly use that token forest from lexer module, and don't need to maintain `SymFrame`s.

There is also a decision, that is, should we traverse AST by token forest, or expanded AST by `read-syntax`? The first is less accurate, but works for incorrect code. The latter is more correct, but requires correct code. And the latter method would be even more simple to write, only need a small function that can be imagined in brain at once.

And also please update the feature matrix in README if necessary. To say which language is supported.

6cdh (changes requested):

doclib/doc.rkt
dannypsnl: What I'm not sure is non s-expression language family.

I thought lexer should be fine with them, but `read-syntax` are all different in these cases?

dannypsnl: Rebuild on lexer token forest.

6cdh: > What I'm not sure is non s-expression language family.

The lexer module uses `module-lexer` in `syntax-color-lib`, that recognizes language and routes to language's own lexer. But it only works for `#lang ...`, not for `#reader ...` and `(module mod-name lang ...)`. It also requires the `#lang lang_name` is correct and recognizable, the rest code can be wrong. For `#reader ...` and `(module mod-name lang ...)`, it fallback to `racket-lexer`.

`read-syntax` always works but requires the code is "readable" by the language's own reader.

doclib/doc.rkt
```diff
@@ -795,6 +805,187 @@
...
+                                         (LexerTokenSpan-end name-span))
+      #:children children))
+
+  (define (definition-symbol-kind head function?)
+    (match (hash-ref *definition-form-kinds* head)
+      ['struct SymbolKind-Struct]
+      ['module SymbolKind-Module]
+      ['function SymbolKind-Function]
+      ['variable SymbolKind-Variable]
+      ['fn-or-var (if function? SymbolKind-Function SymbolKind-Variable)]))
```
6cdh: Can we just define `*definition-form-kinds*` values as `SymbolKind-*`, instead of these intermediate symbols?

doclib/doc.rkt
```diff
@@ -795,6 +805,187 @@
...
+;; definitions up to the enclosing symbol. The forest tolerates incomplete code
+;; (forms left open mid-edit still parse to the end of the document), so the
+;; symbol tree stays stable during editing.
+
+;; Maps the head symbol of a form to how its SymbolKind is chosen. 'fn-or-var
+;; picks Function when the defined name uses the function shorthand
+;; (define (f x) ...) and Variable otherwise.
+(define *definition-form-kinds*
+  (hash "define" 'fn-or-var
+        "define/contract" 'fn-or-var
```
6cdh, line 823: `*definition-form-kinds*` defined here.

6cdh (changes requested):

doclib/doc.rkt
dannypsnl: Have to leave `'fn-or-var`

6cdh (approved):

6cdh:
Commits need to be verified. I recommend you commit manually, or use a password manager to manage SSH keys, it will always popup a menu and ask you if accept when everytime AI try to sign the commit and access the SSH key.

dannypsnl:
No, the problem is `user.signingkey`

Issue #216, opened 2026-06-15
`#lang at-exp racket/base` wrongly be reported as `Missing language declaration`


dannypsnl:
<img width="646" height="188" alt="Image" src="https://github.com/user-attachments/assets/56497467-a3f9-4c69-9bb0-52ecb7328f30" />

6cdh:
I'm tracking this bug.

6cdh:
The bug happens because `doc-lang.rkt` parses `#lang` line using this regex: `#px"^#lang\\s+(\\S+)$"`. It didn't recognize `#lang at-exp racket/base` style language line. But the real root reason comes from two places: the vague behavior of language prefix parsing, and weird new added diagnostics logic.

PR #217, opened 2026-06-17
Redesign doc lang and diagnostics for language header parsing

6cdh:
Solve #216

I did a redesign for language header parsing and diagnostics header check.

Now doc-lang.rkt use a unified terms: parse header and language policy. It has two clean stages:

1. parse the header into a `Language-Header` record.
2. Match it against a predefined `language-specs` list that produces a `Language-Policy` struct that will be used by diagnostics, formatting, expansion, etc.

`rktd` now inside the predefined list define its policy for features, and no need special logic.

I also pass `lexer-state` (that includes `Language-Policy` data) to check-syntax, and services. So they don't need to run lexer again.

Language Header Check in diagnostics module only checks

- language header is missing, or
- malformed language header: header exists, but cannot parse a language name

It does not detect any other issues.

I also improved "collection not found" diagnostics messages that come from reader and expander:

```racket
#lang racke
```

now has this message:

```
Cannot find language "racke".
  module path: racke/lang/reader
Check that the language name is correct and the package is installed.
```

And

```racket
(require foo/bar)
```

now has this message:

```
Cannot find module "foo/bar" in collection "foo".
Check that the module name is correct and the package is installed.
```

### DIFF
--- a/doclib/doc-lang.rkt
+++ b/doclib/doc-lang.rkt
@@ -1,6 +1,7 @@
 #lang racket/base
 
-(require "../common/path-util.rkt"
+(require "../common/interfaces.rkt"
+         "../common/path-util.rkt"
          "lexer/scan.rkt"
          "lexer/snapshot.rkt"
          "lexer/token-tree.rkt"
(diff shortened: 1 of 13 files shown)


doclib/doc-lang.rkt
```diff
@@ -10,28 +10,50 @@
...
+; A plain `#lang` directive.  Inherits `name` from Language-Declaration (the
+; language name, e.g. "racket/base").
+(struct/contract HashLang-Declaration Language-Declaration
+  ()
+  #:transparent)
+
+; A wrapper-shaped `#lang` directive such as `#lang at-exp racket/base`.
+; Inherits `name` from HashLang-Declaration (the wrapper language, e.g.
+; "at-exp"). `delegate-name` records the wrapped language/reader selector.
+(struct/contract WrapperHashLang-Declaration HashLang-Declaration
```
dannypsnl: Can such wrapping be nested?

dannypsnl (dismissed):

doclib/doc-lang.rkt
6cdh: Yes, for example, `#lang errortrace at-exp racket/base`. I think this module has multiple unnecessary or overcomplex concepts and needs a redesign. Don't merge this PR yet.

dannypsnl: So in your model that example became
```
(WrapperHashLang-Declaration
  #:name "errortrace"
  #:delegate-name
  (WrapperHashLang-Declaration
    #:name "at-exp"
    #:delegate-name "racket/base"))
```
but `delegate-name` is a string or `#f`?

6cdh: No, it does not recognize 2 reader wrappers at all. So that code is treated as missing language header... To add this support, it also requires language policy changes. For example, `#lang errortrace racket/base` is same as `racket/base`, but `#lang at-exp racket/base` is `at-exp`, and `#lang errortrace at-exp racket/base` is `at-exp`.

I've redesigned doc-lang module, so the whole module becomes more cleaner, and works for all of these. I will push it soon.

6cdh:
Just hited a [semantic conflict bug](https://martinfowler.com/bliki/SemanticConflict.html). Now everything should be ok.

doclib/service/diagnostic.rkt
```diff
@@ -10,15 +10,19 @@
          data/interval-map
          "../../common/interfaces.rkt"
          "../internal-types.rkt"
-         "../doc-lang.rkt"
+         (only-in "../lexer.rkt"
+                  LexerState-language-policy
+                  Language-Policy-header-range
+                  Language-Policy-header-status
+                  Language-Policy-require-header?)
```
dannypsnl, line 17: Where is this identifier came from?

dannypsnl: I found.

dannypsnl (approved):

PR #218, opened 2026-06-29
Improve json type DSL and allow recursive types

6cdh:
Before:

A bare identifier `Name` can refer to both json types or Racket predicates. It checks if decoder `jsexpr->Name` is bounded, and treat it as a json type if bounded, otherwise treat as Racket predicate. But for recursive types that decoders not yet defined, it treat `Name` as predicate and produce ambiguity bugs.

After:

It does not check anything at static or expand time. So recursive types can work now.

Because a generated decoder can reference other decoders that not defined yet, decoders are generated within lambda wrappers. The generated type contracts and predicates for jsexpr are also generated within lambda wrappers for the same reason.

Other changes:

- Bare identifiers are always treated as json types, except builtin types.
- Add `(contract C)` for Racket contract expression `C`.
- Rename DSL syntax keyword: `hash/c` to `hashof`, `or/c` to `union`.
- Add some "builtin types": `string?`, `boolean?`, `number?`, `exact-integer?`, `exact-nonnegative-integer?`, `symbol?`. they are bare identifiers, but treated as syntax sugars of `(contract string?)`, `(contract boolean?)`, etc.
- Improve error messages for decoders because they are inside anonymous lambda wrappers now.

### DIFF
--- a/tests/json-struct-test.rkt
+++ b/tests/json-struct-test.rkt
@@ -2,10 +2,15 @@
 (require "../common/json-util.rkt" rackunit syntax/macro-testing)
 
 (module+ test
+  (define (capture-fail-exn thunk)
+    (with-handlers ([exn:fail? values])
+      (thunk)
+      (fail "expected an exception")))
+
   ;; Define test structs
   (define-json-struct Pos
-    [line integer?]
-    [char integer? #:json character])
+    [line exact-integer?]
+    [char exact-integer? #:json character])
 
   (define p1 (Pos 1 10))
   (define p2 (Pos #:char 20 #:line 2))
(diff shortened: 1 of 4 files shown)


tests/json-struct-test.rkt
```diff
@@ -2,10 +2,15 @@
...
+    (with-handlers ([exn:fail? values])
+      (thunk)
+      (fail "expected an exception")))
+
   ;; Define test structs
   (define-json-struct Pos
-    [line integer?]
-    [char integer? #:json character])
+    [line exact-integer?]
+    [char exact-integer? #:json character])
```
dannypsnl, line 13: Is here need to wrap with `contract`? Also, should we use non-negative integer here?

6cdh: Don't need. I just changed my statement above to make it more clear. `exact-integer?` is a builtin type, it is a syntax sugar of `(contract exact-integer?)`.

We can use non-negative integer here, but it's just a test. I think don't need to change it.

dannypsnl (approved):

PR #219, opened 2026-07-25
Add same-file hover source detail

6cdh:
relate to #206

Show a same-file source summary for local declarations and their uses. I only concentrate on Racket and Rhombus compatibility. But the code should also work for other s-exp languages.

This is how it works:

- From expand syntax (pre syntax), walk structurally and collect source candidates ranges of nodes that contain given hover position (Rhombus has a specialized logic).
- For each definition target / local arrow start, pick:
  - nearest = smallest enclosing candidate (bottom line of what hover should show)
  - source = largest candidate that starts on the identifier’s line, else nearest
  - display = often that same-line header through line end, else full nearest form
- Scan leading own-line comments above display-start.
- Store ranges in trace service. Query code will combine ranges and declaration service, build the hover string.

For a symbol, its smallest enclosing node (nearest) except itself is the bottom line of what must show. It also tries to show a little bigger context, that is the largest enclosing node that starts at the same line as the symbol, it may or may not ends at the same line. If its end line is more larger than the nearest node's end line, then shows a `...` after nearest node's end line.

It also tries to show comments above the line of the symbol. For functions, those comments are usually its doc comments.

All of them have cap limit, so they don't show too much characters or run too long. And the declaration need to be in the **same file** to display, otherwise, it would show "imported from ...", like before.

I also did a redesign of hover layout.

## Service Design

I also explored how services should grow. Right now trace service exposes raw interval maps, it feels not good. Since this PR, I'd like to gradually migrate to let trace service exposes lower level services, and let lower level services provide basic query methods. And services should not read each other. Let doc.rkt assembly and handle cross-service logics and data.

## Screenshot

Screenshot From 2026-07-26 00-26-06

Screenshot From 2026-07-26 00-26-38

Screenshot From 2026-07-26 00-30-13

Screenshot From 2026-07-26 00-24-22

Screenshot From 2026-07-26 00-24-34

Screenshot From 2026-07-26 00-25-02

It also naturally works for other s-exp languages:

Screenshot From 2026-07-26 00-25-14

And new hover layout:

Screenshot From 2026-07-26 01-07-27


doclib/service/hover/types.rkt
```diff
@@ -0,0 +1,51 @@
...
+         (struct-out Hover-Detail))
+
+;; Own-line comment above the display window.
+;; Ranges are recorded when the trace is built. `doc-hover` reads [start, end)
+;; from the live buffer. `truncated?` is true when the line hit the per-line
+;; character limit.
+(struct/contract Hover-Comment-Line
+  ([start exact-nonnegative-integer?]
+   [end exact-nonnegative-integer?]
+   [truncated? boolean?])
```
dannypsnl, line 27: Which character limit? Do you mean the editing file’s line width (if defined), or the hover displayed message box has width limit?

doclib/service/position-journal.rkt
```diff
@@ -0,0 +1,71 @@
...
+;;
+;; Interval-map keys update on every edit. Positions inside stored values update
+;; only when something reads them. An old snapshot can stay valid without
+;; rewriting every stored value on each keystroke.
+;;
+;; Hover details hold many absolute ranges. Rebuilding them on every edit was
+;; too slow, so we use this log instead.
+;;
+;; Rule: each stored value belongs to one snapshot. Record edits only after that
+;; snapshot is filled. `reset!` clears the log when the maps are cleared.
```
dannypsnl, line 15: When will we invalid this log?

doclib/service/hover/types.rkt
6cdh: All code and comments shown in the hover box. They are defined in code, including:

```
max (shown) source code summary lines (max-hover-source-lines): 10
max source code summary characters (max-hover-source-characters): 1000
max leading comment lines (max-hover-comment-lines): 10
max characters per comment line (max-hover-comment-line-characters): 200
```

For comment lines (above the symbol), it has per line characters limit and max lines limit. For source code summary, it has max line limit and total characters limit. For truncated lines or truncated long comment line, it shows a `...` there. So basically the whole hover box is bounded, won't show too much stuff and hurt editor performance.

doclib/service/position-journal.rkt
6cdh: They almost never become invalid except service `reset`. But remember a new set of service instances are generated every time expansion succeed. So old data will be GCed after each succeed expansion. The edge case is expansion always fail, but it's not likely happen in real life. And position journal only store a list of ranges of change, even if it stores one million items, it wont be a problem. Normally human won't edit that many times. For AI, they usually ensure each batch of edits produces at least correct expandable code.

dannypsnl (approved):

PR #223, opened 2026-08-03
Add workspace model for multi-document and workspace states

6cdh:
## Intent

This PR fits the gap of multi-document analysis states.

- Add a `workspace/state.rkt` that owns workspace and multi-document states. keep `doclib` single document.
- Migrate cross-file references global state from `doclib` to workspace model.
- Do not perform cross-document lookup in doclib.
- Design workspace model well and leave a path for possible future extensions: background expander (indexer), dependency graph, cross file features: hover, rename, workspace symbols, etc.

### Single Document

New Types Prototype:

```racket
;; Unique identifier for cross file symbols
(struct/contract Module-Binding
  ([filepath path?]
   [submods (listof symbol?)]
   [phase+space phase+space-shift?]
   [id symbol?])
  #:transparent)

;; Document Contribution
(struct/contract Doc-Contribution
  ([path path?]
   [references (hash/c Module-Binding? (listof Location?) #:immutable #t)]
   [definitions (hash/c Module-Binding? Location? #:immutable #t)])
  #:transparent)

;; Reference locations returned by one document
(struct/contract Reference-Source
  ([path path?]
   [locations (listof Location?)])
  #:transparent)

;; Live document references plus the exact cross-document identity, when any.
(struct/contract Document-Reference-Result
  ([source Reference-Source?]
   [module-binding (or/c Module-Binding? #f)])
  #:transparent)
```

`Doc` owns two related analysis values:

- the mutable `build-trace%` used by local document features
- an immutable `Document-Contribution` representing the last successfully expanded cross-document state.

A successful trace replaces the old trace, trace version, and the contribution.
Incremental edits may shift the live trace, but not the `Document-Contribution`.

### One source of binding-use data

`declaration%` becomes the only source of truth.

### Workspace ownership

Workspace prototype:

```racket
(make-workspace)

(workspace-add-folder! workspace path)
(workspace-remove-folder! workspace path)
(workspace-contains? workspace path)

(workspace-set-contribution! workspace contribution)
(workspace-remove-path! workspace path)
(workspace-reference-sources workspace module-binding)
(workspace-definition-location workspace module-binding)
```

Workspace stores

- a semaphore for concurrency access
- workspace folders
- contribution per source path. Another Contribution-Store stores them in some optimized data structures, and provides efficient queries on contributions.

The contribution is shared between `Doc` and workspace. This is why it should not be mutable.

Workspace model does not own expand, schedule, opened documents.

A process-wide singleton of workspace instance exists.

### Expansion reuse on open

A future background cache may transfer exclusive ownership of a expand result `CSResult` to an opened document instead of repeating foreground expansion. `CSResult-trace` is mutable, so the ownership transfer is one-shot.

### Background analysis

A future producer can expand closed workspace files and install their contribution values.

### Other cross-file features

When a new feature exists, add a new field to `Doc-Contribution`, and a new workspace query.

EDIT: Replaced the idea section with a more concrete design section.

### DIFF
--- a/common/version.rkt
+++ b/common/version.rkt
@@ -1,11 +1,15 @@
 #lang racket/base
 
-(provide version>=9.0?)
+(provide version>=8.8?
+         version>=9.0?)
 
 (require version/utils)
 
 (define (version>=? target-version)
   (not (version<? (version) target-version)))
+
+(define (version>=8.8?)
+  (version>=? "8.8"))
 
 (define (version>=9.0?)
   (version>=? "9.0"))
--- a/common/version.rkt
+++ b/common/version.rkt
@@ -1,16 +1,15 @@
 #lang racket/base
 
-(provide version>=8.8?
-         version>=9.0?)
-
 (require version/utils)
+
+(provide support/check-syntax-phase-level+space-callback?
+         support/racket/phase+space?)
 
 (define (version>=? target-version)
   (not (version<? (version) target-version)))
 
-(define (version>=8.8?)
+(define support/check-syntax-phase-level+space-callback?
   (version>=? "8.8"))
 
-(define (version>=9.0?)
-  (version>=? "9.0"))
-
+(define support/racket/phase+space?
+  (version>=? "8.3"))
--- a/doclib/doc.rkt
+++ b/doclib/doc.rkt
@@ -31,6 +31,7 @@
          racket/class
          racket/set
          racket/list
+         racket/phase+space
          racket/string
          data/interval-map
          "check-syntax.rkt"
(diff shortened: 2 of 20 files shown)


common/version.rkt
```diff
@@ -1,12 +1,16 @@
...
 (require version/utils)
 
 (define (version>=? target-version)
   (not (version<? (version) target-version)))
 
+(define (version>=8.8?)
+  (version>=? "8.8"))
+
 (define (version>=9.0?)
   (version>=? "9.0"))
```
dannypsnl: Maybe we should define it as

```rkt
(define (phase+space-callbacks?)
  (version>=? "8.8"))
```

and we even can repeat them, e.g.

```rkt
(define (support-xxx-feature?)
  (version>=? "8.8"))
```

rather than a meaningless name here? Because we are anyway care what functionality get supported in which version, not the version itself.

6cdh: It's now improved. `version.rkt` owns a set of feature support predicate. Call sites owns domain rules, for example, composite multiple predicates.

doclib/doc.rkt
```diff
@@ -906,8 +961,10 @@
   (-> Doc? string? Pos? string? (or/c WorkspaceEdit? #f))
   (define-values (start end decl) (doc-get-decl doc pos))
   (match decl
-    [(Decl req? id left right)
-     (cond [req? #f]
+    [(struct* Decl ([filepath filepath]
+                    [left left]
+                    [right right]))
+     (cond [filepath #f]
```
dannypsnl, line 967: `filepath` and `req?`? I'm not sure about this is equivalent.

6cdh: I think they are. It was a local variable rename. It should not break anything.

6cdh:
I refactored `declaration%` service, now it stores definition locations and uses locations separately, provides query methods and export references and definitions contributions. All related document queries use query methods by `declaration%` service, makes their code simpler. And `definition%` is merged to `declaration%`, because it is a small wrapper.

Find references request now uses the new workspace model.

Some extra gains are

- Document highlight now correctly highlight unused definition, and symbols exported by module language (`define`, `+`, etc).
- O(log n) expand/contract methods of `declaration%` service.
- Rename now handle unused definition correctly.
- Slightly faster check-syntax because workspace service is removed.

6cdh:
I didn't migrate definition, hover, rename queries. They requires background expand and other logic polish. That would be easy because the definitions and references contributions are already bridged.

This PR is large enough. I'd like to delay them to next PR. This PR should be complete now.

doclib/service/declaration.rkt
```diff
@@ -1,102 +1,357 @@
...
+    (define module-binding->def-id (make-hash))
+    (define def-id->module-binding (make-hash))
+    (define range-journal (make-position-journal))
+    ;; `'module-lang` uses held until walk-stx.
+    (define pending-module-lang-uses '())
+
+    (define/private (fresh-id!)
+      (define id next-id)
+      (set! next-id (add1 next-id))
+      id)
```
dannypsnl: We can use the following trick:

```rkt
(struct id ())

(define a (id))
(define b (id))
(define c a)

> (eq? a b)
#f
> (equal? a b)
#f
> (eq? a c)
#t
> (equal? a c)
#t
```

eliminate a private state.

6cdh: The `id` struct allocates for each id. Integer id cost less. A private state would be invisible with abstraction. I changed the definition to be

```racket
(define fresh-id!
  (let ([id 0])
    (lambda ()
      (begin0 id
        (set! id (add1 id))))))
```

So the private state is hidden.

dannypsnl (approved):

PR #228, opened 2026-09-04
Hint struct fields at constructor calls

### LINKED ISSUE #29: Struct Field Name Virtual Text Inference (Inlay Hints)
rust-analyzer provides a very neat feature in that it will print out virtual text of struct field names and function param names when you call said function.

This feature would be a huge productivity gain for racket as structs will be much easier to work with, along with functions. Currently I create a lot of keyword functions for structs that are very large because I don't want to keep guessing what they are by position. (I have looked into the rebellion packages and etc but ehh, I prefer to minimize dependencies)

image

```
pub struct Rect {
    pub x: u32,
    pub y: u32,
    pub w: u32,
    pub h: u32,
}
impl Rect {
    pub fn new(x: u32, y: u32, w: u32, h: u32) -> Self {
        Self { x, y, w, h }
    }
}
//so we can have a more ergonomic syntax for creating structs by position
let r = Rect::new(0, 5, 32, 32);
```

dannypsnl:
Every argument of a struct constructor call gets the name of the field it fills, so `(point 1 2)` reads as `(point x 1 y 2)`.

A hint is only produced when the whole field list is known and matches the call. A subtype contributes its supertype's fields first, `#:auto` fields are not arguments, and an argument already written with its own field's name is left alone.

Inlay hints now is a list of inlay-hints-source, configurate for each language variant.

Resolve #29

---

# Demo

image


6cdh:
This is my understanding of the algorithm:

1. Scan token tree for every token list head element that is a candidate constructor O(nodes)
2. Calls Check Syntax's `definition-at` using the candidate's position
3. Search the token forest for the node that contains the definition range O(top level nodes)
4. Check its enclosing form that begins with `struct` or `struct/contract`
5. parse the field list

So it requires expand for inlay hints, and only code included in last successful expand get inlay hint labels.

Therefore we can actually do it in a service by traverse and partial interpret pre-expand syntax objects. This is much more simpler than work on token tree because:

- The pre-expand syntax objects is complete S-expr, and discard insignificant nodes like whitespaces and comments
- Racket has builtin support for work with syntax objects, it's simpler and intuitive than manually traverse and match complex token nodes.

This can simplify the whole algorithm and transform it to a compact interpreter core. And it can also apply to last Typed Racket inlay hint PR. A problem is the service needs to depend on other services' data, but this is easy to workaround.

I forgot it in last PR and suddenly remember we can do it 😂. Actually most features can be done in services by working on pre-expand and post-expand syntax objects. Token trees are only useful for features that are required to work for edited code since last successful expand or need whitespace/comments info, etc.

dannypsnl:
I just make a tried, it's a lot faster (from 1 ~ 3ms to 0.19 ~ 1.2ms).

However, I would like to have another PR for that, because it introduces some new behavior problems when editing the file: token version takes the current text and recompute, syntax is a cache. Hence use pre-syntax can mismatch to the state of what's editing.

It worth to take some efforts to make it works as good as token version when changing files.

6cdh:
You can build data structures using pre-syntax objects positions for efficient query. Then use interval map operations to adjust position data in `expand`/`contract` method for each edit. Another method is using position journal that record edit changes and adjust position data lazily when query happens.

What about this PR? Should we merge now?

6cdh (approved):

dannypsnl:
Yes, I think we can have this PR merged, I will try your suggestions as a standalone fix PR.

jestarray:
Yay! No more scrolling up constantly to double check struct field names, and no more cluttering up code with a million keyword arguments just to know what goes where and keep things readable. It's a big quality of life improvement that makes a huge difference, thanks for uplifting the productivity of everyone guys :partying_face:

PR #230, opened 2026-09-10
Read inlay hints from pre-expand syntax, once per analysis

dannypsnl:
A hint source now reads the pre-expand syntax during the analysis that produced it, collecting the struct declarations as it goes so that second search becomes one hash lookup, and inlay-hint% keeps what it returned for later requests to filter. Each hint records the spans it was read from, so an edit shifts the hints it left alone and drops the ones whose own text it changed, and what stays on screen still describes the text under it even while the document does not expand.


6cdh:
You can use `syntax-parse` to express shapes such as Typed Racket field declaration, `define`, `define-values`, and let clauses more clearly. And use syntax class to simplify form recognition.

## Example

Define a parameterized class for struct declarations:

```racket
(define-syntax-class (struct-declaration struct-head?)
  #:attributes (name super field-list)

  ;; (struct child parent (field ...) option ...)
  (pattern (head:id name-id:id super-id:id fields option ...)
    #:when (struct-head? #'head)
    #:when (syntax->list #'fields)
    #:attr name #'name-id
    #:attr super #'super-id
    #:attr field-list #'fields)

  ;; (struct point (field ...) option ...)
  (pattern (head:id name-id:id fields option ...)
    #:when (struct-head? #'head)
    #:when (syntax->list #'fields)
    #:attr name #'name-id
    #:attr super #f
    #:attr field-list #'fields))
```

The syntax class can be further defined as incomplete `struct` form on need for error tolerance.

Read struct declaration:

```racket
(define (read-declaration stx accepted-heads)
  (define (struct-head? head)
    (memq (syntax-e head) accepted-heads))

  (syntax-parse stx
    [(~var decl (struct-declaration struct-head?))
     (list (syntax-e (attribute decl.name))
           (and (attribute decl.super)
                (syntax-e (attribute decl.super)))
           (syntax->datum (attribute decl.field-list)))]
    [_ #f]))
```

Field specification

```racket
(define-syntax-class racket-field
  #:attributes (value)

  (pattern name:id
    #:attr value (Field-Spec (symbol->string (syntax-e #'name)) #f))

  (pattern (name:id option ...)
    #:attr value
    (Field-Spec
      (symbol->string (syntax-e #'name))
      (for/or ([option (in-list (attribute option))])
        (eq? '#:auto (syntax-e option))))))

(define-syntax-class typed-racket-field
  #:attributes (value)

  (pattern (name:id (~datum :) _type)
    #:attr value
    (Field-Spec (symbol->string (syntax-e #'name)) #f)))
```

Field spec

```racket
(define (racket-field-spec stx)
  (syntax-parse stx
    [field:racket-field (attribute field.value)]
    [_ #f]))

(define (typed-racket-field-spec stx)
  (syntax-parse stx
    [field:typed-racket-field (attribute field.value)]
    [_ #f]))
```

Constructor call:

```racket
(define-syntax-class constructor-call
  #:attributes (head (args 1))

  (pattern (head-id:id args ...+)
    #:do [(define arguments (attribute args))]
    #:when (syntax-start #'head-id)
    #:when (andmap syntax-start arguments)
    #:when (not (ormap syntax-keyword arguments))
    #:attr head #'head-id))

(define (read-call stx)
  (syntax-parse stx
    [call:constructor-call
     (cons (syntax-e (attribute call.head))
           (map syntax->datum (attribute call.args)))]
    [_ #f]))
```

This should be able to simplify code significantly.

doclib/service/inlay-hint.rkt
```diff
@@ -0,0 +1,112 @@
...
+;; Whether an edit changed the text inside `range`. An insert at either end of
+;; a form adds text next to it, not inside it, so only a strictly interior one
+;; counts; a delete counts as soon as it takes any of the form away.
+(define (insert-disturbs? range start _end)
+  (and (< (CharRange-start range) start)
+       (< start (CharRange-end range))))
+
+(define (delete-disturbs? range start end)
+  (and (< start (CharRange-end range))
+       (< (CharRange-start range) end)))
```
6cdh: The claim under test was: *if you change the text a hint was read from, that hint disappears; otherwise it only moves*. That is what `Inlay-Hint-Group` and `features.md` say. The code implements a weaker rule: *if `expand`/`contract` reports a point insert or a deleting interval that hits a source span, drop; otherwise shift*. Those two rules are not the same, because the document does not tell this service "this interval’s text changed."

This is a pitfall of the `expand` and `contract` methods.

The server never sends "replace `[start, end)` with this string" to inlay hints. It only sends a net length change:

https://github.com/jeapostrophe/racket-langserver/blob/d84878a099140c5afaaf70c9593d09e82dd8655a/doclib/doc.rkt#L186-L190

Longer new length means insert the extra characters at the old end. Shorter new length means delete only the leftover characters at the end.

This code was extracted from old interval map range adjustment in each services in a refactor several years ago. And it was only used for range or position adjust before. This is the first time it's used for content change judge. The bugs here are:

**1. Same-length replace keeps stale hints**

`(struct point (x y))` plus `(point 1 2)` shows `x` and `y`. Select the `x` in the field list and type `z`. Old length 1, new length 1. `expand`/`contract` are not called. `groups` is untouched. The call still shows `x`.

**2. A growing replace at a span’s end is treated as "next to the form"**

Source span `[10, 20)`. Replace that whole form with 15 characters (was 10). The trace call is `expand(20, 25)`: insert 5 characters at 20.

Insert only disturbs if the point is strictly inside in `insert-disturbs`. It returns false in this case, and the group is kept and shifted. Every character of the form was overwritten, but the algebra thinks you typed after it.

doclib/service/inlay-hint.rkt
```diff
@@ -0,0 +1,112 @@
...
+    (super-new)
+
+    (define pre-syntax #f)
+    (define groups '())
+
+    ;; The hints drawn in [req-start, req-end], in the order they were read.
+    (define/public (hints-in-range req-start req-end)
+      (for*/list ([group (in-list groups)]
+                  [anchor (in-list (Inlay-Hint-Group-anchors group))]
+                  #:when (<= req-start (Inlay-Hint-Anchor-pos anchor) req-end))
```
6cdh: A hint whose position is `req-end` is returned. It should be excluded because LSP requests use exclusive `end` position.

doclib/service/inlay-hint.rkt
```diff
@@ -0,0 +1,112 @@
...
+             [pos (move-position (Inlay-Hint-Anchor-pos anchor) start end)])))))
+
+(define (edit-groups groups disturbs? move-position move-range start end)
+  (for*/list ([group (in-list groups)]
+              [edited (in-value (edit-group group
+                                            disturbs?
+                                            move-position
+                                            move-range
+                                            start
+                                            end))]
```
6cdh: The each variable stay at each line style formatting feels waste lines. Consider use less lines.

doclib/service/inlay-hint.rkt
```diff
@@ -69,7 +69,8 @@
     (define/public (hints-in-range req-start req-end)
       (for*/list ([group (in-list groups)]
                   [anchor (in-list (Inlay-Hint-Group-anchors group))]
-                  #:when (<= req-start (Inlay-Hint-Anchor-pos anchor) req-end))
+                  #:when (and (<= req-start (Inlay-Hint-Anchor-pos anchor) req-end)
+                              (< req-end (Inlay-Hint-Anchor-pos anchor))))
```
6cdh, line 73: The condition is always false.

Should be

```racket
(<= req-start (Inlay-Hint-Anchor-pos anchor) (sub1 req-end))
```

6cdh (approved):

PR #231, opened 2026-09-14
Add formatter configuration and 2 formatters: DrRacket and fmt

6cdh:
Now we support 3 formatters: fixw, DrRacket and fmt.

- fixw: use its API, it's the default formatter. Only supports Sexp languages.
- DrRacket: DrRacket formatter. It uses a reimplemented class that implements [textoid](https://docs.racket-lang.org/syntax-color/color-textoid.html) interface. DrRacket provides `racket-text%` that also implements `textoid` interface, but it depends on GUI library, that we removed it before. Same as DrRacket, it also supports `drracket:indentation` or `drracket:range-indentation` hooks that can indent customized languages. Supports more languages.
- fmt: runs its CLI `raco fmt` that reads from stdin and print to stdout. Only supports Sexp languages.

The configuration only works for Sexp languages, other languages always use DrRacket.

We have two types of formatting and configurations now:

- full document formatting, for full document formatting request. fixw, DrRacket, fmt all support.
- indentation formatting, for range formatting and on type formatting requests. fixw and DrRacket support, fmt does not.

The configuration details are documented. This change does not require editor client changes.


dannypsnl (approved):

jestarray:
@6cdh Awesome! Do you know where I would set this in vscode? I'm trying `settings.json` but it says `Unknown Configuration Setting` on hover and its greyed out.
image

6cdh:
@jestarray your screenshot is correct. Ignore VSCode's warning. Because it cannot know racket-langserver's configuration format. You can remove `resyntax` or `fmtSettings` section if you do not want to override the default.

jestarray:
@6cdh Ahh ok thanks! Readme should mention to ignore vscode warning. Maybe these settings could be put in magicracket if possible

sidkshatriya:
There seems to be some inconsistency between how the `workspace/didChangeConfiguration` client notification is handled and how the `workspace/configuration` request is handled.

~~I've been investigating this because I can't get the configuration to work in nvim.~~ (I now have got this working in nvim, but the inconsistency is a small bug that nevertheless still needs to be fixed) 

Here is the possible inconsistency:

```racket
(define (fetch-configuration request-client uri)
  (when (client-capability-workspace/configuration?)
    (request-client "workspace/configuration"
                    (->jsexpr
                      (ConfigurationParams
                        #:items (list (ConfigurationItem
                                        #:scopeUri uri
                                        #:section "racket-langserver"))))
                    update-configuration)))

```
**Note that the key 'racket-langserver' is extracted above**

```racket
(define (didChangeConfiguration params)
   (match-define (hash-table ['settings settings]) params)
   (update-configuration settings))
```
**Not that the top `settings` key is extracted above**

I am new to the code base so its a bit confusing to me. Anyways, my LSP configuration for racket in neovim is:
```lua
...
vim.lsp.config('racket_langserver', {
    filetypes = { "racket", "scheme", "rhombus", "scribble" },
    settings = {
        ['racket-langserver'] = {
            formatting = {
                documentFormatter    = "fmt",
                --  indentationFormatter = "fmt" will NOT work as only fixw and drracket are supported here
                indentationFormatter = "drracket",
                fmtSettings          = {
                    width = 50,
                    indent = 2,
                    maxBlankLines = 1
                }
            }
        }
    }
})
```

**Edit:**
- The above neovim LSP config now works ! 
-  `indentationFormatter = "fmt"` will NOT work as only fixw and drracket are supported. This was my earlier mistake and that is why my settings were not being picked up !
- The inconsistency mentioned above is nevertheless a small bug for which I will submit a PR

jryans:
> Maybe these settings could be put in magicracket if possible

I will look around the VS Code + LSP ecosystem to understand the best way for extensions to include langserver settings (so that they can be UI configurable) and attempt to adapt Magic Racket as needed.

PR #236, opened 2026-09-15
Fall back to shipped defaults when didChangeConfiguration has no racket-langserver section

youdie006:
A client that sends `workspace/didChangeConfiguration` with `settings: null`, `settings: {}`, or a settings object holding only another server's section raises `match-define: no matching clause` out of the notification handler, and the stale formatter settings stay in place instead of returning to shipped defaults.

`settings: null` is what `vscode-languageclient` sends on every configuration change when the server uses the pull model, which this server does (`lsp/text-document.rkt:27-35`).

The comment two functions above the break already says so:

> A `racket-langserver` section is a snapshot. Omitted keys use shipped defaults. `workspace/configuration` returns a list; `workspace/didChangeConfiguration` **may send the settings object, `null`, or an empty object**.

`update-configuration` honours all three through `normalize-configuration-item` (`lsp/workspace.rkt:119-133`), and `fetch-configuration` relies on that tolerance for null list items. Only the new unwrapping line in `didChangeConfiguration` hard-fails on the same condition.

Severity, stated honestly: the server does not crash. `workspace/didChangeConfiguration` is dispatched from `handle-notification` (`lsp/methods.rkt:190`), which has no handler of its own, and the outer loop at `main.rkt:75` catches it with `report-error`, which only `eprintf`s (`main.rkt:50-51`). So the observable effect is configuration silently not applied plus stderr noise, not a dead server.

The condition is precisely "the `racket-langserver` key is absent". Extra sibling keys were already fine; I measured that separately.

Why the existing suite did not catch it: `settings-test.rkt` ships a case literally named "null and empty configuration return to shipped defaults", but it calls `update-configuration` directly, so it passes identically before and after #235 and is structurally blind to the wrapper. The two new cases drive the exported `didChangeConfiguration`.

<details>
<summary>Verification</summary>

Racket 9.3 CS in a container with the package installed. `compiled/` cleared before every row, exit codes captured with no pipe in between.

| variant | `lsp/workspace.rkt` md5 | new cases | shipped `settings-test.rkt` |
|---|---|---|---|
| pristine at `c9997cf` | `5e1a2415e6b0385e8468f2ef442f3570` | **exit 3, 3 fail** (null, `{}`, foreign section) | 11/11 |
| this PR | `48cfcade0358f5e335d8f40628a4174a` | **exit 0, 4/4** | 13/13 |
| undo #235 entirely | `859c8d39bef692dd10182720f89c9d85` | exit 1, the "scoped section is applied" case fails | 11/11 |
| `hash-ref settings 'racket-langserver (hasheq)` | `3e76f43d4d101c49af205603872f968f` | exit 1, the `null` case fails | 11/11 |

Four distinct md5s produced four distinct outcomes, so every swap took effect.

The third row matters: it shows these tests pin #235's behaviour rather than undoing it. The fourth shows why the obvious one-liner is not enough, since `settings: null` arrives as `'null` and is not a hash, so `hash-ref` raises.

`raco fixw --ensure-newline-eof .` exits 0 and leaves both files byte-identical, so the formatting job is clean as written.

Not run: the full `raco test -c racket-langserver/tests` suite, and no live client over JSON-RPC. The reproduction calls the real exported `didChangeConfiguration` from the real module, not a stub.

</details>

On signatures: I know this project requires verified signatures, and **my two commits here are unsigned**. I created them through the contents API expecting GitHub to sign them, and it did not, because my credential is a classic token rather than an OAuth app. I did not want to work around that by adding a signing key to my account without thinking it through, so I am flagging it rather than hiding it. The merge commit of #235 has a single parent and is GitHub-verified, so a squash merge would sign this too. If that does not satisfy the requirement, tell me and I will redo the change through the web editor.

Disclosure: written with AI assistance (Claude). Every number above is measured, not predicted.


dannypsnl (dismissed):

sidkshatriya:
Thanks for continuing work on this. I was able to get the formatter working satisfactorily in nvim but then kakoune started giving problems. Here is the diff that I suggest (this is a diff on the current `master` branch commit and not a diff on your PR) as a another solution of the problem you have documented i.e. that some servers may not send the `racket-langserver` key.

This hooks into `normalization-configuration`. It maybe a bit more robust because it deals with lists etc.

If you fused it with your tests it might be better ??

```diff
diff --git a/lsp/workspace.rkt b/lsp/workspace.rkt
index a3fefdb..860c41d 100644
--- a/lsp/workspace.rkt
+++ b/lsp/workspace.rkt
@@ -119,7 +119,9 @@
 (define (normalize-configuration-item item)
   (if (eq? item (json-null))
       (hasheq)
-      item))
+      (match item
+        [(hash-table ['racket-langserver langserver-settings]) langserver-settings]
+        [_ item])))

 (define (update-configuration settings)
   (define normalized
@@ -134,6 +136,5 @@

 (define (didChangeConfiguration params)
   (match-define (hash-table ['settings settings]) params)
-  (match-define (hash-table ['racket-langserver langserver-settings]) settings)
-  (update-configuration langserver-settings))
+  (update-configuration settings))

```

6cdh:
Please sign your commits. This is a requirement made by the repository owner, not a style preference.

For configuration models, I've forgot the details, just did some research. Here are related VS Code and LSP issues: https://github.com/microsoft/vscode-languageserver-node/issues/380 and https://github.com/microsoft/language-server-protocol/issues/1792

VS Code send `{ "settings": null }` as an empty event, means the push model is deprecated, not necessary means the configuration is empty. In this case, LSP server should send `workspace/configuration` request to client to pull configurations. The pull model always send correct data.

So the logic should be, in `didChangeConfiguration` request, if its `settings['racket-langserver']` exists, treat it as valid config and parse it. Otherwise, send `workspace/configuration` request to client.

sidkshatriya:
@6cd our comments were separated by a few seconds. Please do have a look at the alternate solution I suggested above in https://github.com/jeapostrophe/racket-langserver/pull/236#issuecomment-5678182172 . Maybe it is a viable approach ? Thanks !

6cdh:
@sidkshatriya Thanks for your solution. But it didn't solve the root cause. The root cause is, when client (VS Code) send a `{ "settings": null }`, it means configuration should be obtained by using pull model, not means the configuration is empty. While add more error handling code would always make it fallback to the default settings, not read real settings from the client. Your solution recursively search "racket-langserver" section in `settings` field, but it solves a different problem that is `settings` can carry any data and might put `racket-langserver` section in nested JS objects. For this problem, the pull model solves it better, it allows server explicitly request a `racket-langserver` section from the client.

youdie006:
@6cdh You are right and my fix had the wrong semantics. Falling back to shipped defaults treats `settings: null` as "the configuration is empty" when it means "the push model does not apply, ask me". That discards the client's real settings instead of re-reading them. Thanks for the two upstream links, they settle it.

Pushed `172e100`, which does what you described: apply a pushed `racket-langserver` section when there is one, otherwise issue `workspace/configuration`.

Two mechanism decisions I had to make, both easy to change if you would rather they went the other way:

- `fetch-configuration` and the `workspace/configuration` capability move from `text-document.rkt` to `workspace.rkt`, next to `update-configuration`, which they feed. `text-document.rkt` already required `workspace.rkt`, so this removes a cross-module dependency rather than adding one, and it avoids a cycle.
- `ConfigurationItem.scopeUri` becomes `(optional string?)`, which matches the LSP specification where `scopeUri` is optional. That lets the workspace-wide pull omit it. The per-document call at `did-open!` still passes a uri.

The correction is observable, not cosmetic. Running the new tests against the previous version of this branch, the one that fell back to defaults, still fails the pull case:

| `lsp/workspace.rkt` | result |
|---|---|
| master | 2 failures: no pull, and the missing-section case raises |
| this branch's first version, defaults fallback | 1 failure: it resets instead of pulling |
| `172e100` | 14 of 14 pass |

There is a third test asserting that nothing is requested when the client did not advertise the capability, so a client that cannot answer a pull is not left worse off. `raco fixw --ensure-newline-eof .` exits 0 and leaves all five files byte-identical. `tests/lib/doc-test.rkt` reports 8 of 130 failures both with and without this change, so those are pre-existing.

On signing: I cannot produce a GitHub-verified signature from where I work. There is no GPG key here, and the credential I authenticate with is a classic token, so even commits created through the contents API come back `verified: false`, which is why the first two on this branch are unsigned. I am not going to add a signing key to the account just to get a green badge without thinking it through, and I would rather say that than quietly leave it.

So the options as I see them: the merge commit of #235 has a single parent and is GitHub-verified, so a squash merge would sign this too. If that does not satisfy the rule, I am happy to redo the change through the web editor the way #235 was done, or to close this and let you take the diff directly. Your call.

@sidkshatriya Thanks for the alternative. I think 6cdh is right that recursively unwrapping the section in `normalize-configuration-item` solves a different problem, since it still treats an absent section as empty rather than as a request to pull. Your nvim and kakoune report is useful though, since it says clients disagree here in practice.

6cdh (approved):

6cdh:
@youdie006 This repository requires all commits to have verified signatures. The commits in this PR are currently unsigned, so the PR cannot be merged.

If your environment cannot sign commits directly, please either recreate the commits through GitHub's web interface by operate browser, which signs web commits automatically, or ask your host/operator to configure commit signing for you.

Please update the PR once all commits show as Verified.

6cdh:
Thanks!

